PR #7, opened 2023-09-24
Add section for emerging hardware trends

ShvetankPrakash:
Created a placeholder/skeleton for emerging hardware trends that we all can flesh out. (Chapter 19 might not be the best place for this, but didn't want to mess with structure yet so just added it to the end for now.)


_quarto.yml
```diff
@@ -46,6 +46,7 @@ book:
     - responsible_ai.qmd
     - sustainable_ai.qmd
     - generative_ai.qmd
+    - emerging_hw.qmd
```
jveejay, line 49: Could you move this to be after `hw_acceleration.qmd`

jveejay: Take a look at the hw_acceleration contents, there is some stuff in there but I do like the idea of pulling out emerging trends separately, either in the same file or in a new file.

jveejay, line 49: Could you please move this to after the `hw_acecleration.qmd` file so that the hardware components are all together.

jveejay: @mpstewart1 thinks we should just fold this into the "Future Trends section," could you do that. You are welcome to add content in there since that will be separate from what others are doing.

jveejay (changes requested):

_quarto.yml
ShvetankPrakash: Just to clarify: You are saying to rename this to "Future Trends"? I don't see an existing section on the website for future trends to fold this into. And this would then stay at the end as chapter 19 then I'm assuming?

jveejay: Hi @ShvetankPrakash, no that's not what I was suggesting. We were thinking that we should merge your content into the hw_acceleration chapter. In there, there is a future trends section. So move your material in there. So that way it is a nice comprehensive chapter.

uchendui:
Circle CI should stop running tests now.

PR #12, opened 2023-10-01
made changes for assignment (itai shapira's sections)

ishapira1:
please accept this change.

https://github.com/ishapira1/cs249r_book


uchendui:
Thanks for the PR, Itai! Could you please combine these changes with the rest of your group's, and submit a single PR for the chapter?

PR #16, opened 2023-10-05
Add files via upload

Mjrovai:


Mjrovai:
Adding exercises to chapters 2 and 4

PR #25, opened 2023-10-09
Updated Ai4Good chapter

marcozennaro:


ai_for_good.qmd
```diff
@@ -60,6 +70,8 @@ Explanation: Scientific research often requires data collection in challenging e
 - Ecology: Utilizing TinyML for tracking animal migration, studying habitats, and understanding ecological changes.
 - Physics and Chemistry: Enhancing experimental setups with real-time data analysis, aiding in quicker hypothesis testing and validation.
 
+The Ribbit Network was developed to provide accurate readings of carbon dioxide gas through a crowdsourced network of open-source, low- cost, smart sensors [8]. This kind of high-quality data will help scientists better understand and predict the impacts of climate change.
```
profvjreddi, line 73: Could you please update [8] to a real reference, seems like a copy paste issue. 
```suggestion
The Ribbit Network was developed to provide accurate readings of carbon dioxide gas through a crowdsourced network of open-source, low- cost, smart sensors [@???]. This kind of high-quality data will help scientists better understand and predict the impacts of climate change.
```

ai_for_good.qmd
profvjreddi, line 1: These updates are great @marcozennaro! 

One suggestion for improvement would be if we could add some specifics about why or what about tinyML makes the solution tractable compared to more traditional systems. Do you think you could make a pass in general around that?

ai_for_good.qmd
```diff
@@ -1,10 +1,11 @@
...
+By aligning AI progress with human values, goals, and ethics, the ultimate goal of ML systems (at any scale) is to be a technology that reflects human principles and aspirations. Initiatives under "AI for Good" promote the development of AI to tackle the UN Sustainable Development Goals (SDGs) using embedded AI technologies, expanding access to AI education, amongst other things. While it is now clear that AI will be an instrumental part of progress towards the SDGs, its adoption and impact are limited by the immense power consumption, strong connectivity requirements and high costs of cloud-based deployments. TinyML, applowing ML models to run on low-cost and low-power microcontrollers, can circumvent many of these issues.
 
 > The "AI for Good" movement plays a critical role in cultivating a future where an AI-empowered society is more just, sustainable, and prosperous for all of humanity.
 
 ::: {.callout-note collapse="true"}
+
 ## Learning Objectives
 
 * coming soon.
```
profvjreddi, line 11: I will synthesize this as we make progress on it, so don't worry about it for now.

ai_for_good.qmd
```diff
@@ -37,12 +42,14 @@ The portability, lower power requirements, and real-time analytics enabled by Ti
...
 
 - Remote health monitoring: Using TinyML for wearable health devices.
 - Disease prediction and prevention: Early detection systems.
 - Assistive technologies: Devices for the differently-abled.
 
+Examples of TinyML projects in healthcare is a low-cost solution to monitor atrial fibrillation and sinus rhythm developed by UNIFEI in Brazil to extend health coverage in poor regions of the world.  A collaborative paper led by researchers from the University of Khartoum in Sudan and the ICTP  in Italy presents the use of a low-cost device based on TinyML to detect mosquito species, which could lead to a more efficient malaria eradication campaign (@altayeb2022classifying). A prototype rapid water-bone cholera detector kit pluggable into existing taps has been developed at ACEIoT in Rwanda to lower the cost of mass deployment [@ooko2021tinyml). The prediction of water-bone cholera is done by monitoring water's physicochemical patterns. All of these projects support Sustainable Development Goal 3: Good Health and Wellbeing.
```
profvjreddi, line 51: Could we elaborate on what is low-cost cause for folks like you and me we would know this all too well. But for others it might not be so obvious even though they would generally get the idea. For instance, could we add things like this which is a draft of a suggestion: 

> Traditional methods for monitoring heart rhythms like atrial fibrillation and sinus rhythm often rely on expensive medical-grade equipment and visits to healthcare facilities. However, this can be prohibitive for populations with limited access and resources. The solution developed by UNIFEI leverages TinyML and small microcontrollers to create a low-cost, accessible monitoring device. 
>
> Specifically, the device uses an off-the-shelf microcontroller costing only a few dollars along with a pulse sensor that can be produced very cheaply. By minimizing the complexity and cost of components, the device becomes affordable for widespread distribution. The low production costs also enable scaling to large numbers of users. Furthermore, the device is designed to be simple to use, not requiring extensive training or expertise. This increases accessibility for patients in developing regions or remote areas without advanced healthcare infrastructure. Overall, the ultra-low-cost design democratizes access to vital health monitoring tools that can save lives. TinyML enables powerful AI capabilities to be deployed in affordable, accessible ways for underserved communities globally.

I could also work with my MedAI students to integrate additional details around it.

ai_for_good.qmd
```diff
@@ -51,6 +58,9 @@ Explanation: With the global population rising, sustainable and efficient farmin
 - Pest and disease detection: Early warning systems.
 - Sustainable farming practices: Optimizing water and fertilizer use.
 
+Agriculture is the common thread which holds the 17 SDGs together. Investing in the agricultural sector can address not only hunger and malnutrition but also other challenges including poverty; water and energy use; climate change; and unsustainable production and consumption. As a concrete example of TinyML being used in agriculture, researchers at UNIFEI in Brazil developed a low-cost method of detecting diseases in
```
profvjreddi, line 61: Again here, I am wondering why is TinyML the right solution for the problem. Do you think we could follow this sort of a template:

1. Explain what is the problem in the SDG
2. Why we think tinyML is useful
3. How tinyML can uniquely address this problem.

If we shoot for three short paragraphs like that repeatedly then it would make a great connection.

profvjreddi, line 61: It would be interesting for us to mention the FarmBeat project from Microsoft, amongst others.

profvjreddi (changes requested):

PR #124, opened 2023-12-11
Fix two missing references in optimizations.qmd

V0XNIHILI:
See title.


profvjreddi:
I will merge this and do a fix. In general, the format for references is [@citekey]. You can see the differences between [@key] vs. @key

PR #192, opened 2024-05-05
changed word "algorithms" to "models"

BrunoScaglione:
Although models are technically algorithms, in the field of ML the word "algorithm" generally refers to the learning algorithm that spits the model as output.

I think using algorithms as a synonym for models can confuse the reader.


profvjreddi:
Thanks for reading through things carefully.

PR #249, opened 2024-06-04
241 Table rendering in Chapter 7 is messed up

profvjreddi:


profvjreddi:
Fixed the rendering issue with markdown. 

CleanShot 2024-06-03 at 20 53 08@2x

Issue #353, opened 2024-08-15
Student Feedback - Chapter 8


jasonjabbour:
**Chapter Eight - Efficient AI**

- In terms of learning objectives for this chapter, we felt like we generally got most of them! However, we wanted to point out that the first learning objective seemed very similar to the fourth. Also, the last two learning objectives seemed to be less captured--we weren't sure what "nuances" were being referred to in the 6th bullet point, and don't remember any mention of "environment" (as mentioned in the 7th bullet point).
- The formatting seemed a bit inconsistent--for instance, **8.2 The Need for Efficient AI** seemed like it would benefit from the formatting maintained in the next section (**8.3 Efficient Model Architectures**) where terms like "TinyML" or "Edge AI" are bolded with colons prior to the paragraphs explaining them. Overall, we think the chapter would benefit from clearer subtitles and subsections with more bolded/header differentiations. 
- The concept of "Pruning" in **8.4 Efficient Model Compression** has been mentioned a couple times up until this point in the book (and we go on to redefine it in Chapter 9). The constant redefinitions make the book feel fragmented--perhaps when we mention these kind of terms that are mentioned/defined in multiple places, we should include hyperlinks to other sections where it is defined more fully, or a kind of "See Chapter 9 for more information" caption to make reading more cohesive. Another example of this is "FLOPs" (defined in **8.7 Evaluating Models**), which was just mentioned in Chapter 7 but not defined; perhaps we should refer Chapter 7 readers to this section in case they are confused.
- Again, bullet points in the table in section **8.6.1 Numerical Formats** would help readers more easily differentiate lines of text.

[Machine Learning Systems - 8 Efficient AI.pdf](https://github.com/user-attachments/files/15930354/Machine.Learning.Systems.-.8.Efficient.AI.pdf)

_Originally posted by @sgiannuzzi39 in https://github.com/harvard-edge/cs249r_book/discussions/256#discussioncomment-9841790_

jasonjabbour:
- [ ] More consistent formatting 
- [ ] Improved subtitles and subsections
- [ ] Reference previous chapters
- [ ] Reformat 8.6.1 Table

Issue #323, opened 2024-07-31
Wording Improvements


jasonjabbour:
Adjust the wording of sections according to this [document](https://docs.google.com/document/d/1YLI2LEHU34KWJjMZ0Vuy3b7KxXJnZgCJkqtNJ7Wau-w/edit?usp=sharing).

PR #401, opened 2024-08-26
Student Feedback Chapter 9

### LINKED ISSUE #354: Student Feedback - Chapter 9
**Chapter Nine - Model Optimizations**

- We feel like we got all of the learning objectives! Maybe bring the 5th bullet point up one (so it's the fourth) to stay in line with the chapter layout.
-  Again, pruning is defined again in **9.2.1 Pruning**. Like we said with Chapter 8, I would make sure these definitions are cohesive and not as duplicated throughout the chapters.
- We found ourselves getting confused on the formatting in section **8.2 Efficient Model Representation**. For instance, maybe bold subtitles like "Overview" or "Structured Pruning", or make their font bigger so they are differentiated from their subsections (like "Components"). I think especially with these longer sections, it's important to differentiate subsections for readers
- In **8.2 Efficient Model Representation**, "sparse weight matrices" were mentioned several times, the first of which was in Table 9.1. We never felt we got a concrete definition on this term, and would love some more explanation.
- A lot of **9.3.1 The Basics** already feels defined in Chapter 8 - in fact, the table in this section is almost the exact same as the one in Chapter 8.
- We wish **9.3.1 The Basics** went into more detail surrounding "mantissa, exponent, and sign" categorizations (as shown in Figure 9.13).
- A couple of silly typos (all shown in-line in PDF): under Table 9.1, it says "we have exapmles" instead of "we have examples." Under the "Lottery Ticket Hypothesis" section, the second paragraph has the sentence, "They empirically discovered by Jonathan Frankle and Michael Carbin." Section **9.4.3 Kernel Optimizations** says "machine learning models onf resource-constrained devices" instead of (we assume) "on." Finally, in **9.5.3 Hardware Optimization Libraries**, a subtitle reads, "Sparsity (ADD SOME LINKS INTO HERE)" - we imagine that was intended for a later edit.
- Visualizations were great in this chapter!
[Machine Learning Systems - 9 Model Optimizations .pdf](https://github.com/user-attachments/files/15931251/Machine.Learning.Systems.-.9.Model.Optimizations.pdf)

_Originally posted by @sgiannuzzi39 in https://github.com/harvard-edge/cs249r_book/discussions/256#discussioncomment-9842571_

jasonjabbour:
Addressing feedback from #353 and #323.


jasonjabbour:
(Chapter 8 was edited in this PR to provide background information about how floating point numbers are represented. This was an area of confusion for students. But, I think this background information fits better in chapter 8 rather than chapter 9)

PR #430, opened 2024-09-03
Chapter 10 Student Feedback

### LINKED ISSUE #355: Student Feedback - Chapter 10
**Chapter 10 - AI Acceleration**

- In terms of learning objectives for this chapter, we did not feel like we walked away with the fourth bullet point or the sixth bullet point. 
- This was an interesting chapter, but the tone felt like an academic paper. We feel like there is a more accessible way to phrase a lot of the information, especially for students who have not encountered a lot of the vocabulary before. Perhaps also emphasize why this is important for students to be learning.
- There was a lot of text in this chapter, which made it hard for us to fully grasp everything. Maybe find some way to cut down on length--there were a few places in the chapter where bullet points would have sufficed (like in the Disadvantages/Advantages section of **10.3.2 Field-Programmable Gate Arrays (FPGAs), 10.3.3 Digital Signal Processors (DSPs), 10.3.4 GPUs, 10.3.5 CPUs,** etc.). Also, including tables in the place of text is often a better way to get the point across, rather than doing both table and text.
- **10.6 Benchmarking Hardware** feels unnecessary, given that the whole next chapter is about benchmarking. Perhaps include one or the other, but both do not seem necessary. 
- We didn't understand what the table in **10.3.1 Application-Specific Integrated Circuits (ASICs)** was trying to get across, especially the "Notes" column. Some more clarification on this would be great! (Also the table does not have a number or way to reference it)
[Machine Learning Systems - 10 AI Acceleration.pdf](https://github.com/user-attachments/files/16033863/Machine.Learning.Systems.-.10.AI.Acceleration.pdf)

_Originally posted by @sgiannuzzi39 in https://github.com/harvard-edge/cs249r_book/discussions/256#discussioncomment-9907347_

jasonjabbour:
Addresses student feedback from #355


profvjreddi:
Thank you @jasonjabbour for the careful pass.

Issue #447, opened 2024-09-11
In "conclusion.qmd"


BravoBaldo:
In "conclusion.qmd",
 	
- Search "clarity and understanding.." ---> Remove 1 point .... or add another!
- Search "We focus on systems " --> "We focusED on systems "
- There are several occurrences of "ML Ops" and a couple of occurrences in "ops.qmd" but the rest of the book talks about "MLOps" (No space) Please Note: Wikipedia says it's the same!
- Search "It also ensures transparency" I don't understand the subject: "It" who/which
- "and I can't wait to see" maybe less personal...?

The chapter "18 Generative AI" GENERATES a lot of expectations ;)

PR #448, opened 2024-09-11
447 In "conclusion.qmd"

profvjreddi:
https://github.com/harvard-edge/cs249r_book/issues/447


profvjreddi:
Thank you for reading all the way to the end and providing feedback on every chapter. That was amazing! 

If you're available, I would love to talk to you. Please drop me a note at my email, and I'd love to find a time to connect.

PR #528, opened 2024-11-18
Student Feedback Chapter 19

### LINKED ISSUE #411: Student Feedback - Chapter 19
**Chapter 19 - AI for Good**

- Great way to wrap up the book! Easy to understand, cool examples.
- A kind of out-there idea spurred by this chapter -- we've talked about "AI for Good" and some environmental implications of AI, but what are some other ways AI is being used for negative objectives? Maybe before/after this chapter, there could be a chapter delving into how ML is being used for things like weaponry, taking away jobs creative fields, or affecting education. I'm sure there are some cool professors from the Social Studies department who might be able to write about this.
- There was some odd capitalizations and grammar mistakes in this chapter that need a look at. For example, the word "Access" or "Hunger" are continuously capitalized, and there are improper uses of dashes (-). I pointed out the ones I noticed in the PDF.
- In **19.1 Introduction**, the author explains how TinyML could be used to advance the SDG goals. I'm confused as to why we're only looking at TinyML here, as opposed to how other ML systems could help forward these goals. Is it because TinyML is the newest, and therefore most likely to make advancements? If so, please explain that. Otherwise, include how other ML systems can make an impact.
- The second paragraph in **19.6 Disaster Response** restates the same sentence in slightly different ways three times. 
[Chapter 19 - AI For Good.pdf](https://github.com/user-attachments/files/16381420/Chapter.19.-.AI.For.Good.pdf)

_Originally posted by @sgiannuzzi39 in https://github.com/harvard-edge/cs249r_book/discussions/256#discussioncomment-10151493_

jasonjabbour:
Addressing feedback from #411


profvjreddi:
Thank you for your edits, @jasonjabbour. As I review the document, I think we should completely revise the tinyML usage in the book chapter -- another topic, of course. It would be more beneficial to present the content in a general and independent manner, rather than tying it to a specific machine learning use case.

PR #571, opened 2024-12-31
Chapter 18 (old chapter 17) Feedback Revisions

### LINKED ISSUE #410: Student Feedback - Chapter 17
**Chapter 17 - Robust AI**

- First and foremost, this chapter was incredibly long -- nearly double the size of some of the other lengthier chapters in this book. It was so much material that it was hard for us to fully grasp what we were reading. We would recommend cutting down immensely. Some of the ways you can do that is by limiting the number of examples in **17.2 Real-World Examples** (one per section would work fine), and cutting down paragraph size in the bullet point-like text in **17.4.2 Data Poisoning**, **17.4.3 Distribution Shifts**, **17.4.4 Detection and Mitigation,** **17.5 Software Faults,** and **17.6.3 Software-based Fault Injection Tools.** Case studies also felt like too much to add to this already lengthy chapter, so cutting those could be beneficial. 
- Again, Case Study naming/formatting should be standardized across all chapters.
- In sticking with other chapter formats, **17.2 Real-World Examples** should probably be renamed "Historical Precedent." Also, the splitting into Cloud/Edge/Embedded subsections here felt unnecessary, and only worked to make the section longer.
- The figure in **17.3.1 Transient Faults** does not have a number/name assigned to it.
- We felt that including a picture of a slideshow in _Figure 17.13_ was distracting -- perhaps cut or replace with actual image of visualization.
- We didn't understand _Figure 17.15, 17.38_ and _17.39_. 
- "Adversarial Attacks" in **17.4.1 Adversarial Attacks** has already been defined earlier in the book, and does not warrant another definition. Same can be said for "Generative Adversarial Networks" in the same section.
- A lot of **17.4.2 Data Poisoning** is a reiteration of Chapter 12, and therefore can be cut.

_Originally posted by @sgiannuzzi39 in https://github.com/harvard-edge/cs249r_book/discussions/256#discussioncomment-10087259_

jasonjabbour:
Addressing feedback from #410


jasonjabbour:
Section removed from *this* chapter. It is an exact copy of a section in chapter 14.

PR #623, opened 2025-01-17
Summary

hzeljko:
Chapter frameworks.qmdi has been edited.
I reviewed and edited the updated chapter 2: ml_systems.qmd


profvjreddi:
CleanShot 2025-01-17 at 13 37 08@2x

Could you take a look at @hzeljko

hzeljko:
@profvjreddi 
&shy; (soft hyphen): A marker for an optional line break at the point where it is inserted. This means that the word can be split at that point if needed for text wrapping, but it will not be split if it is not necessary.

It is used when you want to manually control where words can break, especially in situations where automatic text wrapping is not ideal or correct, such as with long technical terms, compound words, or words that might disrupt visual balance if not broken at specific points.

It does not appear in the HTML and PDF output.

image

profvjreddi:
Ah OK! Thanks for teaching me that. I had no idea. I realize now that in one other older push I may have undone that, but I will keep this in mind for the future and not "fix" it.

Issue #813, opened 2025-04-17
Improve on-device learning chapter


profvjreddi:
The on-device learning chapter needs to be fully updated, and I've been working on it to make it flow more in line with the ML systems-oriented approach.

PR #820, opened 2025-04-27
813 Improve on-device learning chapter

profvjreddi:
https://github.com/harvard-edge/cs249r_book/issues/813


18jeffreyma:
looking, will revert back shortly

PR #1029, opened 2025-11-04
Update DOI for article in introduction.bib

oamazonasgabriel:
Updates BIB for:

Corrected DOI: 10.1016/0010-4809(75)90009-9.
Computer-based consultations in clinical therapeutics: explanation and rule acquisition capabilities of the MYCIN system

[Volume 8, Issue 4](https://www.sciencedirect.com/journal/computers-and-biomedical-research/vol/8/issue/4), August 1975, Pages 303-320

### DIFF
--- a/quarto/contents/core/introduction/introduction.bib
+++ b/quarto/contents/core/introduction/introduction.bib
@@ -324,10 +324,10 @@
     capabilities of the MYCIN system
   },
   author = {Shortliffe, Edward H.},
-  year = {1976},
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thank you, I use [BetterBib](https://github.com/texworld/betterbib) to update the BibTeX automatically. I will check if its online repo has a different entry than what we are using here. Regardless, thank you for the update! I will merge this in and see what betterbib does.

PR #1040, opened 2025-11-07
Suggest an alternate research article

foundingnimo:
Original link returns a 404. Suggested a replacement


profvjreddi:
Thank you Nimo!

foundingnimo:
@profvjreddi - this somehow doesn't seem to affect the live version - does it need a build step?

PR #1050, opened 2025-11-21
Fix typos in files

didier-durand:
Hi,
just a few more.
Didier


didier-durand:
looking forward to receive it!

PR #1061, opened 2025-12-02
Adding Star History chart at bottom of README.md

didier-durand:
Hi,

Adding star-history chart at bottom of README. It should look like picture below

For all details, see https://www.star-history.com/#harvard-edge/cs249r_book&type=date&legend=top-left

image

Cheers,
Didier

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,3 +365,7 @@ Thanks goes to these wonderful people who have contributed to making this resour
 
 Our goal is to educate 1 million AI systems engineers for the future at the edge of AI.
 </div>
+
+## Star History
+
+[![Star History Chart](https://api.star-history.com/svg?repos=harvard-edge/cs249r_book&type=date&legend=top-left)](https://www.star-history.com/#harvard-edge/cs249r_book&type=date&legend=top-left)


profvjreddi:
Thanks @didier-durand, but do you think we should actually add it? I am OK with it but I am curious to hear why you think we should add it.

Call me crazy, but it's how I work -- this year the goal was 10,000 ⭐, woo hoo we made it! 

Next year the goal is 100k ⭐ -- how -- wanna reach students at schools and help them with it! :) Go team 🙌

profvjreddi:
Yeah, I know it's difficult, and that's exactly why I feel excited about it. If each star were a student learning the material, worldwide, no matter where they are, I believe we should be able to reach them. In the end, its just a fun motivation -- sorta like tracking your workouts only makes you better cause you have a goal, it isn't so much about the stars itself :)

profvjreddi:
That all said, thank you for your wonderful support. It takes a small team to make these things happen, and I am grateful for every bit of help and support. Honestly, the fact that people care motivates me to work on it daily in small pieces.

didier-durand:
That's what I like in OSS too: you find coworkers that you would never meet otherwise

PR #1083, opened 2025-12-19
Basic Windows port for install.sh, documentation updated for example usage.

joeswagson:
Uses a PowerShell script with a custom implemented Write-Ansi function for support with terminals that don't recognize Ansi codes (seemingly everything I use).

### DIFF
--- a/tinytorch/site/extra/install.ps1
+++ b/tinytorch/site/extra/install.ps1
@@ -1,0 +1,298 @@
+# ============================================================================
+# TinyTorch Installer
+# ============================================================================
+#
+# USAGE
+# -----
+#   iwr -useb mlsysbook.ai/tinytorch/install.ps1 | powershell -ExecutionPolicy Bypass
+#
+# ============================================================================
+
+$ErrorActionPreference = "Stop"
+
+# ============================================================================
+# Configuration
+# ============================================================================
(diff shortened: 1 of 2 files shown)


joeswagson:
Accidentally created a pr prematurely, I intend to port as much of the documented process as I can. This is currently just a windows installer script.

Edit: The script is still functional, but there hasn't been anything to ensure the tito module runs smoothly on Windows.

profvjreddi:
Thanks for working on this @joeswagson! Windows support is definitely needed.

One concern we have: with separate install.sh and install.ps1 scripts, we now have two codebases to maintain. Any config changes (repo URL, branch, installation steps) would need updating in both places, and the logic could drift out of sync over time.

Do you think there is an approach that would let us maintain a single installer script that works across platforms? Or do you feel the two-script approach is the best path forward here?

Just wanted to raise this before we commit to maintaining both. Would love to hear your thoughts!

rnjema:
@joeswagson Your concern is indeed valid. So far unless there are any breaking changes to the current interation of the installer Bash script, Git Bash supports all the commands currently used. 

On the flip side, implementing a Python script to handle a more *platform-agnostic* boostrapping of the project is also worthy exploring (*which I suppose is what you are suggestively alluding to this in this case*). Happy hacking ✊

PR #1094, opened 2026-01-01
Security: mitigate shell injection in build-baremetal

RinZ27:
Move artifact name evaluation to `env` context.

Interpolating `inputs.artifact_name` directly into the shell script is unsafe as it allows for potential command injection. By using an intermediate environment variable, the value is handled as data rather than executable code during shell expansion.

### DIFF
--- a/.github/workflows/book-build-baremetal.yml
+++ b/.github/workflows/book-build-baremetal.yml
@@ -1159,10 +1159,11 @@ jobs:
         if: matrix.enabled
         id: build-declaration
         shell: bash
+        env:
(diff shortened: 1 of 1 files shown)


profvjreddi:
By the way @RinCodeForge927, I love acknolwedging anyone who helps any bit, so when i push it live you will see your name appear here: [contrib](https://www.mlsysbook.ai/contents/frontmatter/acknowledgements/acknowledgements.html#sec-acknowledgements-contributors-fe00). If you want to check hte dev build you can see it here [dev](https://harvard-edge.github.io/cs249r_book_dev/)

RinCodeForge927:
Thanks for merging this, Prof. Vijay! Glad I could help strengthen the security for the build workflow. I'm really enjoying reading through the materials and look forward to contributing more to the project soon!

PR #1098, opened 2026-01-02
dynamic gen

Dhyan761:


profvjreddi:
Thank you for this thoughtful contribution! The goal of reducing configuration duplication is excellent, and the shared config approach with a generator script is well-designed. I've been wanting ot do something like this myself for a while.

However, I am currently in the middle of a significant restructuring effort - splitting the book into two separate volumes (Volume I and Volume II). This means:

- Chapter paths are changing from `contents/core/` to `contents/vol1/` and `contents/vol2/`
- The overall book structure and organization is actively evolving
- We'll likely need different configurations for each volume

Because of this timing, merging this PR now would create conflicts that would need immediate rework.

**Would it be okay if we close this PR for now and revisit once the two-volume structure is finalized?** At that point, your approach would be even more valuable since we'd have two volumes worth of configuration to keep in sync.

I really appreciate the effort you put into this and the detailed documentation. Please don't hesitate to contribute again once the restructuring is complete!

Issue #1078, opened 2025-12-13
download instructions for Windows systems


Kobra299:
the download instructions for the tinytorch 

curl -sSL tinytorch.ai/install | bash

does not work on windows based systems so could you please supply ones that will work for windows systems

rnjema:
> the download instructions for the tinytorch
> 
> curl -sSL tinytorch.ai/install | bash
> 
> does not work on windows based systems so could you please supply ones that will work for windows systems

@Kobra299 Understand the frustration, although I have been able to get it working on a Windows using Windows Subsystem for Linux (WSL) with a Ubuntu 24.04 image. Took a bit of hacking, but it works just fine. Alternatively, you can explore Git Bash to setup it up locally. In this regard, checkout a modified Bash script  you can use for setting up [here](https://github.com/rnjema/cs249r_book/blob/fix-tt-win-install/tinytorch/site/extra/install.sh) while the team is taking time to review the proposed additions/modifications. On the other hand, [here](https://github.com/joeswagson/cs249r_book/blob/dev/tinytorch/site/extra/install.ps1) is a proposed Powershell script that one of us has developed for use on Windows.

PR #1105, opened 2026-01-08
Patch tinytorch installer script for Windows-Unix Support

rnjema:
This pull request proposes an enhancement to the installer Bash script, thus, in a way addresses #1078 as well as follows up on #1083 .

First of all the enhancement to this script assumes the user will run it on a UNIX environment, thus, for Windows users in particular, running using Git Bash or WSL should work just fine. As I have noticed, the painpoint in the original `install.sh` script emanates from the differences in setting up virtual environments between UNIX and Windows. 

The proposed modification checks the platform before use to dynamically determine how the virtual environment must be setup, on the other hand, it forces the setup to use the Python executable defined within the scope of the virtual environment.


profvjreddi:
@all-contributors please add @rnjema as a contributor for 🧑‍💻 Code in TinyTorch -- forgot to do this. Thanks @rnjema  for helping me out.

PR #1117, opened 2026-01-19
fix: fix module import in Transformers module test

minhdang26403:
The current tests for Transformers module have wrong import path for Transformers, which lead to test failure. This Pull Request attempts to fix that issue.

### DIFF
--- a/tinytorch/src/13_transformers/13_transformers.py
+++ b/tinytorch/src/13_transformers/13_transformers.py
@@ -46,15 +46,14 @@ Let's get started!
 # %%
 #| export
 import numpy as np
-import math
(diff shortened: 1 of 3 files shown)


profvjreddi:
Great catch @minhdang26403! These test imports were indeed pointing to the wrong path.

The tests were trying to import from `tinytorch.nn.TransformerBlock` but that module doesn't exist. The correct path is `tinytorch.core.transformers.TransformerBlock` (as defined in __init__.py line 102).

This was causing the transformer tests to fail with import errors. Simple fix but important for keeping the test suite running.

Thank you for fixing this!

PR #1172, opened 2026-02-13
fixed google auth allow iframe, slow index.html

kai4avaya:
Hi Vijay,
I actually pushed most of this to dev last week. But I noticed that the updates to the community site did not get propagated to the live site. I'm thinking - do you need to do a rebuild to render the newest updates?

Kai


profvjreddi:
Hi @kai4avaya  -- yea we have to trigger the website puoblish, let me merge this in and publish

PR #1186, opened 2026-02-20
feat: add type hints to 01_tensor source file (#1167)

### LINKED ISSUE #1167: Add type hints for all the class, functions in `tinytorch`
* Ensure type hints  are added for all classes and functions in the codebase. 
* Type hints help catch edge cases early and improve code readability and maintainability.
* Without them, issues are only discovered after failing test cases, which slows down debugging and learning.
* This will also benefit the students working on assignments, as type hints provide guidance and reduce ambiguity.

Additional context:
I am currently completing the assignments by removing the provided solutions using a script and solving them independently.

imuday984:
Resolves #1167 for the tensor module. I have added the type hints directly to the source file (src/01_tensor/01_tensor.py). However, I haven't run the tito module complete build step locally to update the generated package files. Let me know if you'd like me to set up the local environment to run the build, or if a maintainer can trigger it, however i have only added one file if this suits same approach can be applied to similar files too.

### DIFF
--- a/tinytorch/src/01_tensor/01_tensor.py
+++ b/tinytorch/src/01_tensor/01_tensor.py
@@ -60,6 +60,7 @@ Let's get started!
 #| export
 
 import numpy as np
+from typing import Any, Tuple, Union, Optional
 
 # Constants for memory calculations
 BYTES_PER_FLOAT32 = 4  # Standard float32 size in bytes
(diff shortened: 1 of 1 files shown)


tinytorch/src/01_tensor/01_tensor.py
```diff
@@ -317,7 +318,7 @@ def memory_footprint(self):
         """
         return self.data.nbytes
 
-    def __add__(self, other):
+    def __add__(self, other: Union['Tensor', float, int]) -> 'Tensor':
```
harishb00: Here, as per the contract, it should accept tensor, float, int or a numpy array. So, using numpy's typing as mentioned in the previous comment 'TensorLike' will make this simple.

PR #1192, opened 2026-02-21
feat: add type hints to modules 03, 04, and 05 (#1167)

### LINKED ISSUE #1167: Add type hints for all the class, functions in `tinytorch`

imuday984:
Resolves #1167 for modules 03 and 04.

Following the discussion with @harishb00 in my previous PR, I have left 01_tensor and 02_activations for them to push. I have applied the suggested __future__ imports and TypeAlias patterns to these new modules for cleaner code.

Note: I added these directly to the source files (src/) but haven't run the tito module complete build step locally to update the generated package files. Let me know if a maintainer can trigger the build.
thank you.

### DIFF
--- a/tinytorch/src/03_layers/03_layers.py
+++ b/tinytorch/src/03_layers/03_layers.py
@@ -61,6 +61,15 @@ from tinytorch.core.activations import ReLU, Sigmoid  # Module 02 - intelligence
 #| default_exp core.layers
 #| export
 
+from __future__ import annotations
+from typing import Union, Any, List, Optional, Tuple
+from typing_extensions import TypeAlias  # or from typing import TypeAlias if Python 3.10+
+import numpy.typing as npt
+
+# Define the custom types the reviewer suggested
+ArrayLike: TypeAlias = npt.ArrayLike
+TensorLike: TypeAlias = Union[ArrayLike, "Tensor"]
+
 import numpy as np
 
 # Import from TinyTorch package (previous modules must be completed and exported)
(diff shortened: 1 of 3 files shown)


Kabir1618 (approved):

imuday984 (commented):
alright then will work on other files too

tinytorch/src/03_layers/03_layers.py
```diff
@@ -899,32 +922,32 @@ class Sequential:
         >>> params = model.parameters()  # All parameters from all layers
     """
 
-    def __init__(self, *layers):
+    def __init__(self, *layers: Layer) -> None:
```
harishb00: * The contract as per the typehints says that this method supports variable number of `Layer`. So, someone calls it can call like func(layer1, layer2, ...)
* If you look at the implementation, it says that input can also be like
    * func([layer1, layer2]), func((layer1, layer2))
 * We should express this fact in the contract. so that students get a chance to write logic for these edgecases.
 * Something like `def __init__(self, *layers: Union[Layer, list[Layer], tuple[Layer, ...]]) -> None:`
 * Please don't take it just like that. I may be wrong. Think through once if this makes sense. 
 * The intention behind the typehints are it's not just for IDEs but to help students to handle edgecases without us giving much details for every possible edgecase.
 * Suggestions apart, you are doing a good job for the community. Thank you so much. Keep up the good work!

codeCraft-Ritik (commented):
Excellent work on this implementation — very clean and well-organized

PR #1210, opened 2026-03-04
fix: update requires-python to >=3.9 to run with uv

Roldao-Neto:


profvjreddi:
@all-contributors please add @Roldao-Neto as a contributor for 🪲 Bug,Code in TinyTorch

PR #1262, opened 2026-03-20
Fix #1221 suppress expected overflow warning in test

### LINKED ISSUE #1221: Module 02: Overflow in Sigmoid unit test, slightly confusing warning message
This showed up when running the test_unit_sigmoid cell:

Image

That is a bit of a mixed message there. Looking at the test, the assertion holds for the Tensor([-1000,1000]), so the warning for overflow is probably expected. Perhaps it would be a good idea to add a comment saying so?

Same happens in the module integration test below, and in the Aha moment output

asgalon:
Fixes issue #1221 . Replaced slightly confusing warning message about expected overflow in exp on tensor with large numerical values) with message saying that the warning was suppressed. So it does no longer show a scary warning in red, but still acknowledges the overflow happening.

---
*By submitting this PR, you agree to release your contribution under the project's CC-BY-NC-ND license.*

### DIFF
--- a/tinytorch/src/02_activations/02_activations.py
+++ b/tinytorch/src/02_activations/02_activations.py
@@ -38,7 +38,6 @@ By the end of this module, you will:
 
 Let's add intelligence to your tensors!
 """
-
 # %% [markdown]
 """
 ## 📦 Where This Code Lives in the Final Package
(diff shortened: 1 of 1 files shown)


profvjreddi:
@all-contributors please add @asgalon as a contributor for 🪲 Bug,doc in TinyTorch

profvjreddi:
Thanks @asgalon -- apprecaite your help 🙏

PR #1271, opened 2026-03-22
Fix #1268 installscript fails on ubuntu LTS 20.04

### LINKED ISSUE #1268: tinytorch: installscript fails on ubuntu LTS 20.04
See discussion item #1087 

ubuntu 20.04 provides git 2.25.1, which has a bug with processing --sparse command line options.
Additionally, python 3.8 acquired dependency problems with the current requirements.txt configuration.

So the prerequisites in install.sh should be updated to git > 2.25.1 and python >= 3.9

Also, the actual problem was not visible because stdout and stderr were completely suppressed and not recoverable in case of a problem. A better way would be to write all messages to a logfile in the TEMP_DIR and erase that in case of success with a stdout message pointing to the log in case of an error.

asgalon:
See discussion #1087 

Made the following changes:

- removed python3.8 support because of dependency problems. It apparently no longer works without complicating requirements.txt, and probably not even then.
- checking for git > 2.25.1 to avoid --sparse command line args bug.

The problem should affect only very ancient distributions, but since it happened only three months ago it probably will happen again. From my experience, many people have difficulties upgrading their linux distribution after a couple of years. I have observed that they often just reinstall the system from scratch and avoid that for data loss issues. 

I have tested the install script with ubuntu 20.04 installed in a docker container and on a current Mac Book Pro with the following results:

ubuntu 20.04:
```
tinytorch@d8f93717750f:~/projects$ ./install.sh

  Tiny?Torch v0.1.9
  Don't import it. Build it.

Checking prerequisites...
? git version > 2.25.1 required. If you are using ubuntu LTS 20.04, please upgrade to ubuntu LTS 22.04 or clone project and extract tinytorch subdirectory manually.
? Found Python 3.8, but 3.9+ is required
  Install: https://python.org/downloads or 'brew install python' or 'apt install python3.9' or python 3.12

? Missing prerequisites. Please fix the issues above.
```

The "?" show up because of restricted terminal support in the docker container terminal output, nothing to do with the changes.

OSX 26.3.1:
```
La-Bestia:~/projects/test[1016]> /Users/peter/PycharmProjects/cs249r_book/tinytorch/site/extra/install.sh

  Tiny🔥Torch v0.1.9
  Don't import it. Build it.

Checking prerequisites...
✓ Git 2.50.1
✓ Python 3.13 (python3.13)
✓ Python venv module
✓ GitHub reachable

Where would you like to install Tiny🔥Torch?
  Press Enter for default: /Users/peter/projects/test/tinytorch

Install directory [tinytorch]: 

This will create a tinytorch folder here:
  /Users/peter/projects/test/tinytorch

What will be installed:
  - Tiny🔥Torch learning modules
  - Python virtual environment (.venv/)
  - tito CLI tool

Source: harvard-edge/cs249r_book (main branch)

Continue? [Y/n] 

[1/4] Downloading from GitHub...
✓ Downloaded TinyTorch (f00520c)                        
[2/4] Creating Python environment...
✓ Created virtual environment using python3.13
[3/4] Installing dependencies...
✓ Installed dependencies                                
[4/4] Verifying installation...
✓ Verified tito CLI

✓ Tiny🔥Torch installed successfully!

Next steps:

  cd /Users/peter/projects/test/tinytorch
  source .venv/bin/activate
  tito setup

Then start building:

  tito module start 01

Documentation: https://tinytorch.ai
```

### DIFF
--- a/tinytorch/site/extra/install.sh
+++ b/tinytorch/site/extra/install.sh
@@ -157,7 +157,7 @@ fetch_latest_version() {
             local tag_name
             tag_name=$(echo "$response" | grep -o "\"name\": *\"${TAG_PREFIX}[^\"]*\"" | head -1 | sed 's/.*"name": *"\([^"]*\)".*/\1/')
             if [ -n "$tag_name" ]; then
-                TINYTORCH_VERSION="${tag_name#$TAG_PREFIX}"
+                TINYTORCH_VERSION="${tag_name#"$TAG_PREFIX"}"
                 return 0
             fi
         fi
@@ -191,7 +191,7 @@ check_python_version() {
     minor=$($python_cmd -c "import sys; print(sys.version_info.minor)" 2>/dev/null)
 
     # Check for Python 3.8+ (Required for TinyTorch)
-    if [ "$major" -eq 3 ] && [ "$minor" -ge 8 ]; then
+    if [ "$major" -eq 3 ] && [ "$minor" -ge 9 ]; then
         echo "$version"
         return 0
     elif [ "$major" -gt 3 ]; then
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thanks @asgalon -- @all-contributors please add @asgalon as a contributor for 🪲 Bug,Code in TinyTorch

profvjreddi:
@all-contributors please add @asgalon as a contributor for Bug,Code in TinyTorch

PR #1321, opened 2026-04-11
fix: export DecisionLog and FailureBanner from mlsysim.labs

Shashank-Tripathi-07:
## Problem

`DecisionLog` and `FailureBanner` are defined in `mlsysim/labs/components.py` and imported directly by 14+ lab notebooks, but were missing from `mlsysim/labs/__init__.py`.

Any lab using the public API:
```python
from mlsysim.labs import DecisionLog
from mlsysim.labs import FailureBanner
```
raises `ImportError` at runtime.

## Fix

Add both to the import block and `__all__` in `mlsysim/labs/__init__.py`.

## Verification

```python
from mlsysim.labs import DecisionLog, FailureBanner  # ✓
```

### DIFF
--- a/mlsysim/labs/__init__.py
+++ b/mlsysim/labs/__init__.py
@@ -24,6 +24,8 @@ from .components import (
     RooflineVisualizer,
     LatencyWaterfall,
     MathPeek,
+    DecisionLog,
+    FailureBanner,
 )
 
 __all__ = [
@@ -43,4 +45,6 @@ __all__ = [
     "RooflineVisualizer",
     "LatencyWaterfall",
     "MathPeek",
+    "DecisionLog",
+    "FailureBanner",
 ]


Shashank-Tripathi-07:
## How this breaks the lab workflow

Every lab notebook starts with a setup cell that imports from `mlsysim.labs`. The Design Ledger pattern used across all labs looks like this:

```python
from mlsysim.labs.components import DecisionLog
from mlsysim.labs.state import DesignLedger
```

Because `DecisionLog` and `FailureBanner` are missing from `__init__.py`, this import works fine (direct submodule path), but the intended public API fails:

```python
from mlsysim.labs import DecisionLog  # ImportError
```

### Impact

**DecisionLog** is used in 14 lab notebooks (vol1: lab_00, lab_01; vol2: labs 05–16). It renders the persistent architectural decision log at the end of each lab — the component that carries student decisions forward across the entire curriculum. If it fails to import, the entire lab cell errors out on load and the student sees a broken notebook before doing anything.

**FailureBanner** is used in lab_05 (Neural Computation). It renders the structured failure feedback UI when a prediction lock is wrong — core to the predict→discover→explain loop that the labs are built around.

Both components are already fully implemented in `components.py` — this is purely a missing re-export in `__init__.py`.

profvjreddi:
Thank you @Shashank-Tripathi-07. I've been spreading myself a bit thin between working on the mlsysim and the labs, so some things are slipping through the cracks so I appreciate you catching this.

@all-contributors please add @Shashank-Tripathi-07 for code,bug in MLSysIM

PR #1335, opened 2026-04-16
fix(tinytorch): ensure model params have requires_grad=True in trainer_init

### LINKED ISSUE #1334: [TinyTorch] Potential issue in 08_training test_unit_trainer_optimizer_update not updating parameters - v0.1.9
### Module

08 Training

### Type of Improvement

Bug fix

### Description

When running 08_training.py, `test_unit_trainer_optimizer_update()` fails with the assertion `assert changed, "Parameters should change after optimizer update"`. The root cause is that `param.grad` is None in `opti, because the Tensors in SimpleModel's Linear layers were not defined with`requires_grad = True`.

---

Potential confounding factor: I started with v0.1.3, and moved progress to a different computer and resumed with v0.1.9. I apologize if the issue is caused by this.

### Proposed Solution

Force gradients calculation in `trainer_init()` for the model

```python
    # History tracking
    self.history = {
        'train_loss': [],
        'eval_loss': [],
        'learning_rates': []
    }

    # Enable gradient tracking for all model parameters
    # This is needed because layers may be created without requires_grad=True
    for param in model.parameters():
        param.requires_grad = True

    ### END SOLUTION
```

This fixed the issue for me, and allowed 08_training.py to complete.

Shashank-Tripathi-07:
Fixes #1334 — `test_unit_trainer_optimizer_update` fails with "Parameters should change after optimizer update" because `param.grad` is never populated.

**Root cause:** `Linear.__init__` creates weights with `requires_grad=False`. `tracked_mse_forward` only attaches a backward graph `if predictions.requires_grad` — so `loss.backward()` returns immediately and no gradients flow. `Optimizer.__init__` does set the flag but only on construction; any ordering edge case silently breaks the chain.

**Fix:** iterate `model.parameters()` in `trainer_init` and set `requires_grad = True`, making the Trainer the authoritative owner of gradient tracking.

| File | Change |
|------|--------|
| `tinytorch/src/08_training/08_training.py` | `requires_grad = True` loop in solution block; APPROACH docstring updated |

### DIFF
--- a/tinytorch/src/08_training/08_training.py
+++ b/tinytorch/src/08_training/08_training.py
@@ -632,8 +632,10 @@ def trainer_init(self, model, optimizer, loss_fn, scheduler=None, grad_clip_norm
     # Enable gradient tracking for all model parameters.
     # Layers (e.g. Linear) may be created without requires_grad=True,
     # so we set it explicitly here to ensure backward() populates param.grad.
+    # Guard against raw numpy arrays passed by test stubs or non-Tensor params.
     for param in model.parameters():
-        param.requires_grad = True
+        if isinstance(param, Tensor):
+            param.requires_grad = True
 
     # Training state
     self.epoch = 0
--- a/tinytorch/src/08_training/08_training.py
+++ b/tinytorch/src/08_training/08_training.py
@@ -608,8 +608,9 @@ def trainer_init(self, model, optimizer, loss_fn, scheduler=None, grad_clip_norm
 
     APPROACH:
     1. Store model, optimizer, loss_fn, scheduler, and grad_clip_norm as attributes
-    2. Initialize epoch=0, step=0, training_mode=True
-    3. Create history dict with keys: 'train_loss', 'eval_loss', 'learning_rates'
+    2. Enable gradient tracking: set requires_grad=True on all model.parameters()
+    3. Initialize epoch=0, step=0, training_mode=True
+    4. Create history dict with keys: 'train_loss', 'eval_loss', 'learning_rates'
        (each mapping to an empty list)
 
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
## CI Stage 2 Failure — Not caused by this PR

The Stage 2 failure is a pre-existing repo-wide bug in `tests/10_tokenization/test_10_tokenization_progressive.py::TestRegressionPrevention::test_progressive_stability`, unrelated to this fix.

**Root cause:** The test creates a `DummyModel` whose `parameters()` returns a raw `np.array` instead of a `Tensor`:

```python
class DummyModel:
    def parameters(self):
        return [np.array([1.0])]  # ← raw numpy, not a Tensor
```

This was latent before — `trainer_init` never iterated `model.parameters()`. Our fix does, which surfaced the crash:

```
AttributeError: 'numpy.ndarray' object has no attribute 'requires_grad'
```

**Fix already applied:** The `requires_grad` loop now guards with `isinstance(param, Tensor)`, so raw numpy arrays are skipped silently. This is included in the latest commit on this branch. Stage 2 should pass on the next CI run.

profvjreddi:
thanks, merged

@all-contributors please add @Shashank-Tripathi-07 as a contributor for bug, code in TinyTorch

Shashank-Tripathi-07:
Prof. I'm already a contributor to TinyTorch and hence, can't be added again 😄

profvjreddi:
Haha. It's OK I like adding you all multiple times hahahaha :)

Shashank-Tripathi-07:
It's an honour sir, I've also proposed an idea in discussions section which I think you'll be interested 😄

PR #1336, opened 2026-04-16
test(tinytorch): add finite-difference gradient correctness tests for Module 06

Shashank-Tripathi-07:
## Summary

No backward pass in TinyTorch was previously verified numerically. Existing tests only checked that `param.grad` was non-None or non-zero. A subtly wrong gradient would train silently in the wrong direction.

This PR adds a dedicated `test_gradient_correctness.py` using central finite differences as ground truth:

```
df/dx ~= (f(x + e) - f(x - e)) / (2e)
```

If the analytical gradient from `backward()` disagrees, the test fails.

## Coverage

| Category | Tests |
|----------|-------|
| Arithmetic | add, sub, mul, div, matmul (both operands), broadcast add, chained ops |
| Activations | ReLU (incl. zero boundary), Sigmoid, Tanh, GELU |
| Losses | MSELoss, MSELoss batch, BinaryCrossEntropyLoss, CrossEntropyLoss |
| Composed graphs | Linear weight grad, Linear bias grad, two-layer chain, MSE-through-Linear end-to-end |
| Accumulation | gradient accumulates correctly across two backward calls |
| no_grad() | disables tracking inside context, does not affect outside |

## Files

| File | Change |
|------|--------|
| `tinytorch/tests/06_autograd/test_gradient_correctness.py` | New: 393 lines, 32 tests |

### DIFF
--- a/tinytorch/tests/06_autograd/test_gradient_correctness.py
+++ b/tinytorch/tests/06_autograd/test_gradient_correctness.py
@@ -8,10 +8,10 @@ using finite differences as ground truth.
 The core idea: for any function f, the analytical gradient computed by
 backward() should match the numerical gradient:
 
-    ∂f/∂x ≈ (f(x + ε) - f(x - ε)) / (2ε)
+    df/dx = (f(x + e) - f(x - e)) / (2e)
 
-If they disagree, the backward implementation is wrong — silently producing
-incorrect updates during training.
+If they disagree, the backward implementation is wrong and training silently
+learns in the wrong direction.
 
 Coverage:
 - Arithmetic ops: add, sub, mul, div, matmul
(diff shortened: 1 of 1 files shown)


profvjreddi:
@Shashank-Tripathi-07 wanted to close the loop on this one.

github auto flipped this to merged when #1338 landed (your branch tip was reachable through it). then #1340 removed the file from dev to keep ci green. net: you're credited but the file is gone from dev. confusing flow, sorry about that.

your tests actually caught real stuff though:
- tanh is straight up missing from enable_autograd()s patch list. filed #1341 for that, one small PR.
- a bunch of other ops have the graph building but the backward values dont match finite diff. filed #1342 to audit each one.

both are real findings, solid work. if you want to pick up #1341 (mechanical fix, good shape for a followup PR) or grab an op from #1342 (investigative), both are open.

Shashank-Tripathi-07:
Sure, Sir I'll take a look and do the required changes. Thanks for helping out 😄

PR #1338, opened 2026-04-16
fix(tests/10_tokenization): replace raw numpy array params with Tensor in DummyModel

Shashank-Tripathi-07:
## Summary

\`DummyModel.parameters()\` in \`test_progressive_stability\` was returning \`[np.array([1.0])]\`. Any \`Trainer.__init__\` that iterates \`model.parameters()\` and accesses \`param.requires_grad\` crashes with:

\`\`\`
AttributeError: 'numpy.ndarray' object has no attribute 'requires_grad'
\`\`\`

This is a latent bug in the test itself: every other test stub in this file already uses \`Tensor(np.array(...), requires_grad=True)\` for parameters. \`DummyModel\` was the one outlier.

## Fix

Give \`DummyModel\` a proper \`Tensor\` parameter with \`requires_grad=True\`, consistent with the rest of the test file and with what \`Trainer\` actually expects.

## Files

| File | Change |
|------|--------|
| \`tinytorch/tests/10_tokenization/test_10_tokenization_progressive.py\` | \`DummyModel\` now stores a \`Tensor\` param instead of a raw \`np.array\` |

### DIFF
--- a/tinytorch/tests/06_autograd/test_gradient_correctness.py
+++ b/tinytorch/tests/06_autograd/test_gradient_correctness.py
@@ -8,10 +8,10 @@ using finite differences as ground truth.
 The core idea: for any function f, the analytical gradient computed by
 backward() should match the numerical gradient:
 
-    ∂f/∂x ≈ (f(x + ε) - f(x - ε)) / (2ε)
+    df/dx = (f(x + e) - f(x - e)) / (2e)
 
-If they disagree, the backward implementation is wrong — silently producing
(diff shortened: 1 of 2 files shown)


profvjreddi:
thanks, merged

btw, split the stray gradient test file into #1340 to get ci green. keeping those tests in #1336 where theyre being worked through. thanks again

@all-contributors please add @Shashank-Tripathi-07 as a contributor for bug, tests in TinyTorch

PR #1339, opened 2026-04-16
fix(labs/lab01): unblock prediction widget chain

### LINKED ISSUE #1317: [Bug] Labs lab01
### Area

Other

### Location

Labs Vol1 lab01_ml_intro

### Description

- part[ABC]_prediction ui radio elements are defined, but never rendered, so the values are never set. 

### Expected Behavior

Working through the lab until the complete flow seems correct.

_No response_

Shashank-Tripathi-07:
## Problem

Each `mo.stop()` gate cell was returning its interactive widgets (`partA_scenario`/`partA_fix`, `partB_batch`, `partC_target`) bundled with the next prediction radio. When a prediction was `None`, `mo.stop()` halted execution and those returns never happened, breaking the entire downstream chain. Result: `partB_prediction`, `partC_prediction`, `partD_prediction` were never created, so the tab-building cell never ran and nothing rendered.

## Fix

Matches `lab_02`'s working pattern:
- Each gate cell now returns **only** the next prediction radio
- All interactive widgets (`partA_scenario`, `partA_fix`, `partB_batch`, `partC_target`, `partD_scale`) move into the final gate cell (gated on `partD_prediction`)
- Widgets are created only after all four predictions are resolved, then the tab cell can reference them normally

## Testing

Verified against `lab_02_ml_systems.py` which uses the identical pattern and works correctly.

Fixes #1317

### DIFF
--- a/labs/vol1/lab_01_ml_intro.py
+++ b/labs/vol1/lab_01_ml_intro.py
@@ -280,6 +280,57 @@ def _(
 def _(mo, partA_prediction):
     mo.stop(partA_prediction.value is None, mo.md("**Make your prediction above to unlock this part.**"))
 
+    # Part B widgets
+    partB_prediction = mo.ui.radio(
+        options={
+            "A) Data loading (D_vol/BW) -- weight loading dominates at batch=1": "data",
+            "B) Compute (O/R_peak) -- it is a GPU, compute must dominate":       "compute",
+            "C) Framework overhead -- kernel dispatch is the bottleneck":         "overhead",
+            "D) All three terms are roughly equal":                                "balanced",
+        },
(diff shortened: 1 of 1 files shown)


profvjreddi:
reopening to get fresh ci with the workflow fix from #1344

Shashank-Tripathi-07:
Prof. can you explain me why the CI build is failing. I can't get a good hold of that. Your wisdom will help me route this PR better. @profvjreddi

Based on the new ci workflow fix, I've understood it prof. thanks 😄 

even I was thinking that there's something wrong with the build and you were pretty fast to find and correct it 🚀

profvjreddi:
thanks, merged. this matches the lab_02 structure with each prediction radio in its own non-gated cell.

@all-contributors please add @Shashank-Tripathi-07 as a contributor for bug, code in Labs

Issue #1342, opened 2026-04-16
backward class math audit — finite diff mismatch


profvjreddi:
so the finite diff tests from #1336 flagged some value mismatches. did a code read of every Backward class that came up and the math is actually right. so this is narrower than it first looked.

audit results (reading math against textbook derivatives):
- AddBackward, SubBackward, MulBackward, DivBackward, MatmulBackward: math correct
- ReLUBackward, SigmoidBackward, GELUBackward: math correct
- _reduce_broadcast_grad helper: logic correct

so the "value mismatch" failures are most likely:
1. tolerance too tight for legitimate numerical noise. RTOL=1e-3 + EPS=1e-4 may need to be looser per-op, especially for GELU which chains multiple nonlinearities
2. something subtle in the test helper, like how `loss.backward(np.ones_like(loss.data))` propagates through `.sum()` for 0-d outputs

what needs to happen:
1. run one failing test at a time with analytical vs numerical printed side by side
2. if they match within ~1e-2, tune rtol/atol for that op
3. if they differ by more than that, look at the test helper or graph traversal, not the Backward class

note: tanh bypass was a separate real bug, fixed in #1343 and tracked in #1341.

profvjreddi:
updating the scope above after a code review pass. every Backward class that came up here has correct math against the textbook derivative. the real issue is tolerance tuning or something in the test setup, not the Backward classes. saves anyone picking this up from chasing the wrong thing.

profvjreddi:
Closing this out — the audit is complete and there's nothing remaining:

  1. **Backward class math**: All verified correct (Add, Sub, Mul, Div, Matmul, ReLU, Sigmoid, GELU, `_reduce_broadcast_grad`)
  2. **Tanh bug**: Fixed separately in #1343
  3. **Finite diff tests**: The tests from #1336 that flagged mismatches were on the `test/autograd-gradient-correctness` branch, which has been merged and deleted. The permanent test suite
  (`tests/06_autograd/`) uses analytical expected values with `np.allclose` defaults — no finite-difference checker exists in the current codebase, so there are no tolerance-sensitive tests left to tune.

  If a numerical gradient checker is added in the future, per-op tolerance tuning (especially for GELU) should be considered at that time.

PR #1343, opened 2026-04-16
fix(tinytorch): wire Tanh into enable_autograd()

### LINKED ISSUE #1341: tanh missing from enable_autograd() wiring
just noticed tanh doesnt track gradients. if i do Tanh()(x) with x.requires_grad=True, x.grad is None after backward().

looked at enable_autograd() in src/06_autograd/06_autograd.py ... sigmoid, relu, gelu, softmax, mse, bce and crossentropy are all in the patch list but tanh is missing. also no TanhBackward class anywhere.

came up while looking at #1336.

fix is pretty mechanical:
- add TanhBackward following the same shape as ReLUBackward. math is d/dx tanh(x) = 1 - tanh(x)^2
- wrap Tanh.forward in enable_autograd() same way as tracked_sigmoid_forward
- add Tanh to the import line in the try block next to the other activations

profvjreddi:
closes #1341

## what

Tanh was the only activation not patched by `enable_autograd()`. calling `Tanh()(x).backward()` silently left `x.grad` as None because the forward never attached a computation graph.

## changes

follows the existing `tracked_sigmoid_forward` pattern precisely (same "save output for backward" shape since `d/dx tanh(x) = 1 - tanh(x)^2` can be derived from the forward result):

| change | location |
|---|---|
| new `TanhBackward(Function)` class | `src/06_autograd/06_autograd.py` (between SigmoidBackward and SoftmaxBackward) |
| new `tracked_tanh_forward` wrapper | inside `enable_autograd()` |
| Tanh added to activations import + patch install | inside `enable_autograd()` |
| regression test | `tests/06_autograd/test_06_autograd_progressive.py` |

## test

the new `test_tanh_activation_gradient` asserts:
- `x.grad is not None` after calling `tanh(x).sum().backward()`
- values match the analytical derivative at 0 and ±1 within rtol=1e-5

## context

came out of #1336 where @Shashank-Tripathi-07's finite-diff tests exposed the bypass. see #1341 for the diagnosis and #1342 for the followup backward-math audit across other ops.

### DIFF
--- a/tinytorch/src/06_autograd/06_autograd.py
+++ b/tinytorch/src/06_autograd/06_autograd.py
@@ -1734,6 +1734,65 @@ class SigmoidBackward(Function):
         ### END SOLUTION
 
 
+# %% nbgrader={"grade": false, "grade_id": "tanh-backward", "solution": true}
+#| export
+class TanhBackward(Function):
+    """
+    Gradient computation for tanh activation.
+
+    Tanh: tanh(x) = (eˣ - e⁻ˣ) / (eˣ + e⁻ˣ)
+    Derivative: tanh'(x) = 1 - tanh(x)²
+    """
+
+    def __init__(self, input_tensor, output_tensor):
+        """
+        Initialize with both input and output.
+
+        Args:
+            input_tensor: Original input to tanh
+            output_tensor: Output of tanh (saves recomputation)
+        """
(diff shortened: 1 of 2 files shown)


profvjreddi:
credit to @Shashank-Tripathi-07 since this gap was surfaced by his finite diff tests in #1336

PR #1389, opened 2026-04-17
fix(lab05): resolve silent WASM hang on Pyodide boot (#1388)

### LINKED ISSUE #1388: [Bug] vol2/lab_05_dist_train: silent Pyodide hang in WASM export — no widgets render
## summary

found during a local end-to-end audit of all 33 labs: `labs/vol2/lab_05_dist_train.py` exports fine, boots Pyodide without any console error, and its marimo shell attaches — but the interactive tabs / sliders / radios NEVER render. the lab is effectively non-functional in WASM mode.

## repro

1. `cd mlsysim && python3 -m build --wheel`
2. Copy the wheel to `/tmp/wasm/wheels/`.
3. `marimo export html-wasm labs/vol2/lab_05_dist_train.py -o /tmp/wasm/lab_05/index.html --mode run --no-show-code`
4. Serve `/tmp/wasm` with COEP/COOP headers (see `labs/tests/browser_smoke.py`).
5. Open `/lab_05/index.html` in headless Chromium via Playwright.
6. Wait 2 minutes. `[role=\"tab\"]` count stays at 0.

## diagnostics

- `marimo check vol2/lab_05_dist_train.py` → exit 0
- `test_static.py` + `test_engine.py` (both via `app.run()`) → pass
- `browser_smoke.py` → passes with 'Pyodide settled in 5.0s (captured errors: 0)'
- local DOM probe after 2 min: `tabs=0 radios=0 sliders=0`
- console: no `pageerror`, no `console.error`, no `{\"type\":\"exception\"}` payload
- load sequence identical to a working lab (lab_02_ml_systems) up to 'Loaded rich'
- kernel messages diverge: lab_02 emits ~84, lab_05 stops at ~66 — suggests execution stalls after a specific cell

## not caused by recent PRs

the WASM bootstrap in this file is identical to the state after #1353 (6d56fff7a, 'import plotly.subplots AFTER micropip install'). PR #1386 only reordered its return tuples alphabetically; it did not change the bootstrap or any cell logic.

## what to investigate

- which cell between #66 and #67 (approximately) stalls. add `print` markers or use marimo's `--mode edit` export to get cell-level logs.
- whether pint, plotly.subplots, or one of the `mlsysim.core.defaults` imports blocks in the WASM runtime for this lab specifically.
- whether the big tabs cell (~1400 lines in a single function) hits any Pyodide recursion limit when marimo analyses it.

## why existing guards missed this

- static / engine / marimo check all run in native Python, never touch Pyodide.
- the #1374 browser smoke only checks for exceptions + shell attach. a silent `await` hang throws no error and the shell element attaches anyway.
- CI ran green on this lab in #1374 and #1386 for the same reason.

## recommended follow-up guard

extend `labs/tests/browser_smoke.py` to assert a minimum `[role=\"tab\"]` count on labs that declare `mo.ui.tabs(...)`. any lab where the tabs element fails to render after a 60s wait should fail CI.

found this while running a full 33-lab audit. the other 32 labs render correctly.

Shashank-Tripathi-07:
## Summary

Fixes #1388 — `lab_05_dist_train.py` exports cleanly and Pyodide boots without error, but interactive tabs never render in the browser.

- **Root cause**: Cell 0 imported `INFINIBAND_NDR_BW_GBS` from `mlsysim.core.defaults` but the return tuple referenced `IB_NDR_BW_GBS` — a name that was never assigned. This `NameError` causes Pyodide to stall silently mid-execution with no console error, leaving all `mo.ui.tabs` unrendered (~66 kernel messages instead of ~84).
- **Dead imports removed**: `GPU_MTTF_HOURS`, `IB_NDR_LATENCY_US`, `SCALING_EFF_256GPU`, `OVERHEAD_PIPELINE_BUBBLE`, and unused `EDGE` variable were all imported/assigned but never consumed by any downstream cell.
- **Magic numbers replaced**: Hardware tier dropdowns and synthesis cell hardcoded `989.0`, `312.0`, `25.0`, `12.5`, `65.0` — these now use `H100_TFLOPS_FP16`, `A100_TFLOPS_FP16`, `T4_TFLOPS_FP16`, `IB_NDR_BW_GBS`, `IB_HDR_BW_GBS`, `ETH_100G_BW_GBS` sourced from the mlsysim registry.
- **Regression guard**: `browser_smoke.py` extended with Phase 4 — after network-idle, checks that `[role="tab"]` elements are visible for any lab that declares `mo.ui.tabs`. This catches the exact failure mode in #1388 that previously passed all smoke checks silently.

## Test plan

- [x] Export `lab_05_dist_train.py` with `marimo export html-wasm` and verify tabs render in headless Chromium
- [x] Run `browser_smoke.py` against the exported lab — Phase 4 should pass cleanly
- [x] Introduce the original bug (remove `IB_NDR_BW_GBS` alias) and confirm Phase 4 emits `[tabs-missing]`
- [x] Verify other vol2 labs unaffected (no new params injected into their cells)

### DIFF
--- a/labs/tests/browser_smoke.py
+++ b/labs/tests/browser_smoke.py
@@ -115,7 +115,7 @@ def _extract_python_exception(text: str) -> str | None:
     return f"{exc_type}: {msg}"
 
 
-def verify_lab(page, name: str, url: str) -> list[str]:
+def verify_lab(page, name: str, url: str, labs_dir: Path) -> list[str]:
     """Navigate to a lab, let Pyodide boot, then report any captured errors."""
     errors: list[str] = []
 
@@ -174,6 +174,33 @@ def verify_lab(page, name: str, url: str) -> list[str]:
     # to emit its errors to the console before we tally.
     page.wait_for_timeout(int(POST_IDLE_SETTLE_S * 1000))
 
+    # Phase 4: for labs that declare mo.ui.tabs(...), verify that at least one
+    # [role="tab"] element is visible in the DOM. Network-idle passes even when
+    # Python execution stalls mid-cell (as in #1388), because Pyodide downloaded
+    # all wheels — it just never finished running the cells. A rendered tab is
+    # proof that the tabs cell actually executed to completion.
+    tab_elements = page.query_selector_all('[role="tab"]')
+    if tab_elements:
+        # tabs were found — check they are actually visible, not just attached
+        visible_tabs = [t for t in tab_elements if t.is_visible()]
+        if not visible_tabs:
+            errors.append(
+                f"[tabs-hidden] {len(tab_elements)} [role='tab'] elements "
+                f"attached but none visible — tabs cell likely stalled"
+            )
+    else:
+        # No tabs in DOM at all — only flag if the lab source declares mo.ui.tabs
+        # (avoids false positives on simpler labs that have no tabs).
+        lab_source_dir = labs_dir / name
+        source_files = list(lab_source_dir.glob("*.py"))
+        if source_files:
+            lab_src = source_files[0].read_text(encoding="utf-8", errors="ignore")
+            if "mo.ui.tabs" in lab_src:
+                errors.append(
+                    "[tabs-missing] lab declares mo.ui.tabs but no [role='tab'] "
+                    "elements rendered — Python execution likely hung before tabs cell"
+                )
+
     elapsed = time.monotonic() - pyodide_start
     print(
         f"  ✅ {name}: Pyodide settled in {elapsed:.1f}s "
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @Shashank-Tripathi-07. Both things I really appreciated here: (1) the lab05 cleanup actually pulls the constants from the registry instead of hard-coding them, which is the pattern we want everywhere, so thanks for keepign that in check; and (2) the new [role="tab"] visibility check in browser_smoke is a much better signal than network-idle alone — that was the exact failure mode in #1388. This will catch a whole class of silent Pyodide stalls going forward.

The labs are still very much a work in progress, so please keep that in mind.

@all-contributors please add @Shashank-Tripathi-07 as a contributor for 🪲 Bug,Code in Labs

profvjreddi:
@all-contributors please add @Shashank-Tripathi-07 for bug, code in Labs

PR #1392, opened 2026-04-17
test(tinytorch): PyTorch-compat test coverage for Tensor API additions

### LINKED ISSUE #1298: [TinyTorch]
### Module

01 Tensor

### Type of Improvement

Other

### Description

I’ve been working with TinyTorch and had a few ideas that could make it feel more intuitive and closer to a PyTorch-like experience, especially when building ML models.

## 1. PyTorch-like API Aliases

To make the API more familiar and easier to use:

- It might be helpful to add `@property` support for:
  - `ndim`
  - `numel` (to align with common tensor usage patterns)

- Adding `view()` as an alias for `reshape()` would also improve consistency with PyTorch.

- A simple `contiguous()` pass-through method could be useful too, mainly to avoid attribute errors when running training pipelines that expect it.

## 2. More Robust Tensor Initialization

The `__init__` method could be made a bit more flexible when handling different types of inputs:

- For example, if a list of Tensors is passed in, it could recursively extract their `.data`.

- It would also be nice if `requires_grad` is preserved automatically when initializing from an existing Tensor.

## 3. Advanced Indexing Utility

- It could be useful to add a `masked_fill(mask, value)` method, since this comes up often in transformer models .

Shashank-Tripathi-07:
## Summary

Adds `TestTensorPyTorchCompat`, 10 focused tests covering the PyTorch-compatible Tensor methods that landed in #1391 and the subsequent merge window (view, masked_fill, Tensor stacking).

Also syncs `src/01_tensor/01_tensor.py` with the exported package: `view()`, `masked_fill()`, and Tensor stacking were present in `core/tensor.py` but missing from the source file, causing CI failures.

Closes #1298.

**Why these tests matter:** students porting PyTorch training loops to TinyTorch will call these methods constantly. Without explicit coverage, a refactor could silently break the behavioral contracts. The -inf attention-mask pattern in `masked_fill` is especially tricky: it produces no exception, just wrong softmax output.

Each test targets one specific contract:

| Test | What it pins |
|---|---|
| `test_ndim` | ndim matches len(shape) for rank 0-3 |
| `test_numel` | numel() == .size |
| `test_view_same_as_reshape` | view() is a reshape alias |
| `test_view_with_minus_one` | -1 inference passes through |
| `test_contiguous_returns_correct_data` | values preserved |
| `test_contiguous_is_c_contiguous` | C-order layout guaranteed |
| `test_masked_fill_basic` | True positions replaced |
| `test_masked_fill_attention_pattern` | -inf pattern for transformers |
| `test_masked_fill_does_not_mutate_original` | no in-place side effects |
| `test_tensor_from_tensor_list` | Tensor([t1, t2]) stacking |

## Test plan

- [x] `pytest tinytorch/tests/01_tensor/test_tensor_core.py::TestTensorPyTorchCompat -v` all 10 pass
- [x] Full test file runs clean with no regressions in existing test classes

### DIFF
--- a/tinytorch/src/01_tensor/01_tensor.py
+++ b/tinytorch/src/01_tensor/01_tensor.py
@@ -291,6 +291,8 @@ class Tensor:
         HINT: Use np.array(data, dtype=np.float32) to convert data to NumPy array
         """
         ### BEGIN SOLUTION
+        if isinstance(data, (list, tuple)) and len(data) > 0 and isinstance(data[0], Tensor):
+            data = np.stack([t.data for t in data])
         self.data = np.array(data, dtype=np.float32)
         self.shape = self.data.shape
         self.size = self.data.size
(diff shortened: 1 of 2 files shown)


profvjreddi:
Hey @Shashank-Tripathi-07 — thanks so much for this! 🙏 Going to land this one first so the tests are in before anything else touches Tensor.

@all-contributors please add @Shashank-Tripathi-07 for code, test in TinyTorch

PR #1395, opened 2026-04-18
fix(epub): stop sidenote filter from injecting LaTeX into EPUB output

### LINKED ISSUE #1333: [Bug] epub: Side notes are embedded into the text
### Area

Book — Volume I

### Location

ML Systems, TinyML: Ubiquitous Sensing

### Description

The side notes (or footnotes?) are displayed inline in the text, this happens at different places. Noted with the calibre ebook reader and an Onyx Boox reader:

Image

Look for e.g. "1 milliwatt**The 1 mW Threshold**"

Onyx Boox reader example:

Image

Look for "model weights**Model Weights:**"

### Expected Behavior

The explanations should probably be linked as footnotes

_No response_

octo-patch:
Fixes #1333

## Problem

The `sidenote.lua` filter matched on `epub` alongside `latex`/`pdf`:

```lua
if quarto.doc.is_format("latex") or quarto.doc.is_format("pdf") or quarto.doc.is_format("epub") then
```

Because ePub uses HTML internally, the `pandoc.RawInline('latex', ...)` nodes that wrap `\sidenote{…}` are silently dropped by the EPUB renderer. The footnote body (between the delimiters) is emitted inline, causing sidenote/footnote text to appear embedded in the running prose rather than as proper footnotes.

Example from the issue report: `"1 milliwatt**The 1 mW Threshold**"` — where `The 1 mW Threshold` is sidenote content that leaked into the main paragraph.

## Solution

Remove `epub` from the format guard. For EPUB output, `return nil` lets Pandoc's default footnote renderer take over, placing footnote content correctly at the end of the chapter.

```lua
-- Before
if quarto.doc.is_format("latex") or quarto.doc.is_format("pdf") or quarto.doc.is_format("epub") then

-- After
if quarto.doc.is_format("latex") or quarto.doc.is_format("pdf") then
```

## Testing

The fix can be verified by building the EPUB output and checking that footnotes no longer appear inline in the body text. The PDF/LaTeX sidenote behavior is unchanged.

### DIFF
--- a/book/quarto/filters/sidenote.lua
+++ b/book/quarto/filters/sidenote.lua
@@ -1,6 +1,10 @@
 function Note (note)
-  -- Only process for latex, pdf, or epub formats
-  if quarto.doc.is_format("latex") or quarto.doc.is_format("pdf") or quarto.doc.is_format("epub") then
+  -- Only convert footnotes to sidenotes for PDF/LaTeX output.
+  -- ePub is HTML-based: pandoc.RawInline('latex', ...) nodes are ignored by
+  -- the EPUB renderer, so the surrounding \sidenote{} delimiters are stripped
+  -- while the note body is emitted inline — causing the sidenote text to
+  -- appear embedded in the running prose (see issue #1333).
+  if quarto.doc.is_format("latex") or quarto.doc.is_format("pdf") then
     -- For PDF/LaTeX, convert footnote to sidenote for margin placement
     -- Requires sidenotes package (loaded in header-includes.tex)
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thanks @octo-patch 🙏  

The EPUB build has been quietly leaking raw \sidenote{} markup into the prose for a while, and pulling EPUB out of the LaTeX-only path is exactly the right fix. The EPUB build has a few other issues that we need to track down and if you happen to have the time I'd greatly appreciate it! Thanks for taking the time to write up hte comment as well.

@all-contributors please add @octo-patch as a contributor for Bug in Book

PR #1415, opened 2026-04-20
fix(staffml/gauntlet): add rubric checkboxes and fix silent fail on empty question pool

Shashank-Tripathi-07:
## What is fixed

Two bugs in the Gauntlet page.

### Bug 1: No rubric in Gauntlet

The Practice page shows rubric checkboxes after revealing an answer. You check off which points you covered, and the score buttons cap at your rubric score. Gauntlet had none of this. Users were scoring themselves on gut feel with no structure.

This matters because Gauntlet is the mock interview mode, the highest-stakes session in the app. It had weaker self-assessment tooling than Practice.

What this PR adds to Gauntlet:
- After revealing an answer, rubric checkpoints are extracted from the model answer (same logic as Practice)
- Checkboxes appear above the score buttons
- Score buttons use `effectiveMaxScore` (capped by rubric) so buttons above your rubric score are disabled
- The score label changes from "Rate yourself" to "Confirm or override" when a rubric is present
- Rubric resets when moving to the next question

### Bug 2: Silent fail when question pool is empty

In design mode, if no L5+ questions exist for the selected track, `startGauntlet` hit `if (selected.length === 0) return` and did nothing. The user clicked Begin and nothing happened. No feedback, no error.

Now shows a toast: "No questions found. Try a different track or difficulty level."

## How to verify

**Rubric:**
1. Start a Gauntlet session
2. Reveal the answer on any question
3. Rubric checkboxes should appear above the score buttons
4. Check/uncheck boxes and confirm the score buttons cap accordingly

**Silent fail:**
1. This is hard to trigger in the UI since the pool is large, but the code path is now guarded with a toast instead of a silent return

### DIFF
--- a/interviews/staffml/src/app/gauntlet/page.tsx
+++ b/interviews/staffml/src/app/gauntlet/page.tsx
@@ -16,6 +16,8 @@ import { getLevelDef } from "@/lib/levels";
 import { buildReportUrl } from "@/lib/issue-url";
 import { track } from "@/lib/analytics";
 import { saveAttempt, saveGauntletResult, AttemptRecord, recordActivity, updateSRCard } from "@/lib/progress";
+import { extractRubric, rubricToScore, RubricItem } from "@/lib/rubric";
+import { useToast } from "@/components/Toast";
 import NapkinMathDisplay from "@/components/NapkinMathDisplay";
 import QuestionFeedback from "@/components/QuestionFeedback";
 import HardwareRef from "@/components/HardwareRef";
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
The auto-label workflow failed on this PR when it first opened because `ai-action/setup-ollama` could not be downloaded from the GitHub API (transient fetch error). As a contributor I don't have permission to retrigger `workflow_dispatch` on the repo, so this PR ended up without any labels.

profvjreddi:
Thanks so much @Shashank-Tripathi-07! Both of these had been nagging at me 🙏

@all-contributors please add @Shashank-Tripathi-07 for bug, code in StaffML

PR #1530, opened 2026-04-25
fix(mlsysim): hierarchical AllReduce broadcast 8x overestimate + H200 capacity unit mismatch

Shashank-Tripathi-07:
## Summary

Two correctness bugs in `mlsysim/core`, both producing silent wrong numbers:

### Bug 1 -- Hierarchical AllReduce broadcast phase 8x too slow (`formulas.py`)

`calc_hierarchical_allreduce_time()` models NCCL's 3-phase algorithm:
1. Intra-node reduce-scatter: `M` bytes → each GPU holds `M/G`
2. Inter-node AllReduce across nodes: `M/G` bytes
3. Intra-node broadcast back: `M/G` bytes ← **bug here**

Phase 3 was reusing `t_reduce` (computed on full `M`) via a `# Symmetry assumption` comment. But the broadcast only carries `M/G` bytes -- the reduced shard, not the full gradient. `reduced_message` was already computed for phase 2 and just needed to be passed to phase 3.

**Impact:** On an 8-GPU node, broadcast time overestimated by **8x**. For a 16B param model (~32 GB gradients at FP16) on 2×8 H100s, this overcounts ~220 µs per AllReduce step -- every multi-node DP training simulation reports pessimistically slow communication.

```python
# Before (wrong):
t_broadcast = t_reduce  # Symmetry assumption

# After (correct):
t_broadcast = calc_ring_allreduce_time(reduced_message, gpus_per_node, intra_node_bw, intra_node_lat)
```

### Bug 2 -- H200 memory capacity in decimal GB, everything else in binary GiB (`constants.py`)

```python
# Before:
H200_MEM_CAPACITY = 141 * GB   # decimal, 141×10⁹ bytes

# After:
H200_MEM_CAPACITY = 131 * GiB  # 141 GB converted to GiB, matches all other GPU constants
```

Every other GPU (V100, A100, H100, B200) uses `GiB`. The outlier caused ~7% overcounting of H200 memory capacity in cross-GPU comparisons and memory-fit calculations.

## Test plan

- [x] `uv run --with pytest python -m pytest tests/` -- 363 passed, 2 skipped, 0 failures

### DIFF
--- a/mlsysim/mlsysim/core/constants.py
+++ b/mlsysim/mlsysim/core/constants.py
@@ -41,7 +41,7 @@ H100_TDP = 700 * watt                     # SXM variant
 # NVIDIA H200 (Hopper, 2023) — Source: NVIDIA H200 Data Sheet
 # H200 shares the Hopper compute die with H100, only memory differs
 H200_MEM_BW = 4.8 * TB / second             # HBM3e
-H200_MEM_CAPACITY = 141 * GB               # NVIDIA specifies H200 as 141 GB (decimal)
+H200_MEM_CAPACITY = 131 * GiB              # 141 GB (NVIDIA spec) converted to GiB for consistency with other GPU constants
 H200_TDP = 700 * watt                       # Same as H100 SXM
 
 # NVIDIA B100/B200 (Blackwell, 2024) — Source: NVIDIA Blackwell Architecture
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @Shashank-Tripathi-07. Sixth correctness fix from you across mlsysim/tinytorch/staffml — all the same flavour. Keep them coming.

Shashank-Tripathi-07:
Yes sir, a lot of these fixes are needed to finely correct the things. I'm currently focusing on this patchwork till the website release and I'll switch to some big work to make the repo better and do some project-wide contributions. @profvjreddi

PR #1550, opened 2026-04-26
fix(book): give AlexNet figure full float placement options for PDF

### LINKED ISSUE #1423: [Bug] A figure of AlexNet in the book doesn't print well in PDF.
### Area

Book — Volume I

### Location

1.4 Bitter Lesson (Page no. 8 in the book) 

### Description

Image

I was going through the book, revisiting things when I spotted that the AlexNet architecture is not printed well. I've tried changing the pdf readers to see if this was a reandering issue but the issue stays consistent. 

### Expected Behavior

The whole image should've printed correctly that the learner could use to understand Alexnet 

_No response_

profvjreddi:
Closes #1423.

## Summary

Reporter (#1423) observed the AlexNet architecture figure (@fig-alexnet, page 8 / Bitter Lesson section) renders cropped in the Vol1 PDF — only the bottom edge of the figure appears at the top of the page, with the rest cut off. Issue persists across PDF readers, so it is a real LaTeX placement issue.

## Root cause

The TikZ figure is wide (~26.7 cm clip box) and ~7.45 cm tall. With `fig-pos="htb"`, LaTeX is restricted to here/top/bottom of the *current* page. When preceding text exhausts those positions, the figure gets pushed but has no acceptable landing slot, producing the truncated render.

## Fix

Change `fig-pos="htb"` → `fig-pos="!htbp"`:

- **`p`** — float to own page (figures-only page) where the figure has the full text-block height to render.
- **`!`** — override strict placement parameters (totalnumber, topfraction) so the figure can land wherever it physically fits.

Other figures in the book that use the same pattern have not reported this issue, so the targeted change is appropriate without a global sweep.

## Test plan

- [ ] Render Vol1 PDF after merge, navigate to Bitter Lesson section, confirm AlexNet figure renders fully (either at original location or floated to its own page).
- [ ] If still cropped: follow-up fix is to add `out-width="100%"` to the div, or apply `scale=0.85` at the `\begin{tikzpicture}` level.


profvjreddi:
Closing per author guidance — `fig-pos` should not change on this figure. Will pursue a different approach for #1423.

PR #1591, opened 2026-04-27
fix(staffml): correct cloud-0013 INT4 throughput math and unit error in distractor

Shashank-Tripathi-07:
## Bug

`cloud-0013` (The TPOT Memory Wall) presents an INT4-quantized 70B model but the `napkin_math` field showed FP16 arithmetic:

- **Before (wrong):** implied ~24 tokens/sec (FP16: 140 GB / 3.35 TB/s)
- **After (correct):** ~96 tokens/sec (INT4: 70B x 0.5 bytes = 35 GB, 35 GB / 3.35 TB/s = 10.4 ms/token)

Separately, distractor option 1 contained a 1000x unit error: \"3.35 GB/s\" should be \"3.35 TB/s\".

## Changes

- `interviews/vault/questions/cloud/cloud-0013.yaml`
  - `napkin_math`: complete INT4 calculation showing 35 GB model size and ~96 tok/s throughput
  - Option 1 (distractor): fixed unit \"GB/s\" -> \"TB/s\"
  - Option 2 (correct answer): updated from \"~24 tokens/sec\" to \"~96 tokens/sec for this INT4 model\"
  - `correct_index` unchanged (still 2)

**Question text and scenario are untouched.** Only the answer explanation and option text were corrected to match the precision specified in the scenario.

## Verification

INT4 math: 70B params x 0.5 bytes/param = 35 GB. H100 HBM bandwidth = 3.35 TB/s. Theoretical throughput = 3.35e12 / 35e9 = 95.7 tok/s, comfortably above the 20 tok/s requirement stated in the scenario.

**Related:** PR #1590 (cloud-0024 correct_index fix) is a companion StaffML audit fix.

PR #1592, opened 2026-04-27
fix(staffml): correct mobile-0534 napkin_math numbers to match scenario

Shashank-Tripathi-07:
## Bug

`mobile-0534` (The Mobile Style Transfer Jank) has a napkin_math that uses completely different numbers than the scenario:

| | Scenario (correct) | napkin_math (wrong) |
|--|--|--|
| Compute | 200 GOps | 500 MFLOPs (1000x smaller) |
| Data moved | 500 MB | 10 MB (50x smaller) |
| Calculated AI | 400 Ops/Byte | 50 FLOPs/Byte |

The correct option at `correct_index: 1` already says "Arithmetic Intensity (400 Ops/Byte)", which is correct per the scenario. The napkin_math was just never updated to match.

## Fix

`interviews/vault/questions/mobile/mobile-0534.yaml`

Updated `napkin_math` to use the scenario's actual numbers: 200 GOps / 500 MB = 400 Ops/Byte, then compare to A17 Pro ridge point of ~683 Ops/Byte to confirm memory-bound diagnosis.

**Question text, scenario, options, and correct_index are all untouched.**

**Related:** PR #1591 (cloud-0013 INT4 math fix) and PR #1590 (cloud-0024 correct_index fix) are companion StaffML audit fixes from the same audit pass.

### DIFF
--- a/interviews/vault/questions/mobile/mobile-0534.yaml
+++ b/interviews/vault/questions/mobile/mobile-0534.yaml
@@ -27,10 +27,11 @@ details:
     slow model must be compute-bound. They might also confuse memory capacity (total GB of RAM) with memory
     bandwidth (GB/s), which is the actual bottleneck for data-hungry layers. The key is the *ratio* of
     compute to memory access, known as Arithmetic Intensity.
-  napkin_math: 1. **Calculate Arithmetic Intensity (AI):** Assume the style transfer layer performs 500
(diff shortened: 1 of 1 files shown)


profvjreddi:
@all-contributors please add @Shashank-Tripathi-07 as a contributor for 🪲 Bug in StaffML

profvjreddi:
I thought I had poured thorugh this so many times, even manually auditing several of them 😮‍💨

PR #1593, opened 2026-04-28
Fix staffml imports when only summary corpus is present

farhan523:
```md
## Summary
This fixes a local contributor/dev failure in `interviews/staffml` where some code still imports `src/data/corpus.json` even though this checkout only includes `src/data/corpus-summary.json`.

## What changed
- switch `src/lib/taxonomy.ts` to import `corpus-summary.json`
- make the static full-detail fallback in `src/lib/corpus.ts` stop hard-requiring the missing `corpus.json` artifact

## Problem solved
On a fresh checkout, running the staffml interview app can fail with:

```text
Module not found: Can't resolve '../data/corpus.json'
```

This PR aligns those imports with the summary-only bundle currently committed in the repo, so local dev can start without the legacy full-corpus JSON file.

## Existing behavior impact
- no change to summary metadata already provided by `corpus-summary.json`
- local fallback paths no longer crash when `corpus.json` is absent
- behavior remains consistent with the newer summary-plus-worker architecture already described in the code comments

## Question 
Should `src/data/corpus.json` still be treated as a required generated artifact for contributors, or is the summary-only flow now the intended default? If the former, I can follow up with a docs or generation-step fix instead.

### DIFF
--- a/interviews/staffml/src/lib/corpus.ts
+++ b/interviews/staffml/src/lib/corpus.ts
@@ -428,6 +428,7 @@ const VAULT_API = process.env.NEXT_PUBLIC_VAULT_API
 // In-memory cache for hydrated questions during one session.
 const _detailsCache = new Map<string, Question>();
 let _staticDetailsCache: Map<string, Question> | null = null;
+let _staticDetailsUnavailable = false;
 
 function shouldUseStaticDetails(): boolean {
   if (process.env.NEXT_PUBLIC_VAULT_FALLBACK?.toLowerCase() === "static") return true;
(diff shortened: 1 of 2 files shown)


profvjreddi:
Hi @farhan523 — thanks for the fix! This PR currently has merge conflicts against `dev`. Could you rebase on the latest `dev`? I pushed some changes recently that may have already touched these import paths, so it's worth checking whether the conflict resolution still leaves a meaningful diff (or whether the issue is already resolved upstream). Also happy to hear your thoughts on the question you raised about whether `corpus.json` should still be treated as a required generated artifact.

farhan523:
Hi Professor @profvjreddi ,  It looks like the original import-path issue is already resolved upstream, so there isn’t really a meaningful remaining diff for taxonomy.ts.

PR #1594, opened 2026-04-28
Clean up StaffML analytics visibility listener

farhan523:
## Summary
This fixes a small lifecycle bug in interviews/staffml/src/components/Providers.tsx.

## What changed
- replace the inline visibilitychange listener with a named handler
- remove both visibilitychange and pagehide listeners in the effect cleanup

## Problem solved
Providers was registering visibilitychange with an inline callback and only removing the pagehide listener on cleanup. During remounts or local HMR, that can leave duplicate visibility listeners behind and trigger repeated lushAnalytics() calls for a single visibility change.

## Impact
- no UI change
- keeps analytics flush behavior the same in normal use
- avoids listener accumulation across remounts/hot reloads

### DIFF
--- a/interviews/staffml/src/components/Providers.tsx
+++ b/interviews/staffml/src/components/Providers.tsx
@@ -25,11 +25,15 @@ export default function Providers({ children }: { children: React.ReactNode }) {
 
     // Flush pending analytics on page unload
     const handleUnload = () => flushAnalytics();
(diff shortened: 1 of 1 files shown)


profvjreddi:
@all-contributors please add @farhan523 for code in staffml

profvjreddi:
Thanks @farhan523

PR #1599, opened 2026-04-29
fix(instructors): keep theme toggle visible across breakpoints

farhan523:
## Summary
- Exclude the Quarto color-scheme (theme) toggle from the mobile sidebar-hide rule so the dark/light switch stays accessible on small screens.
- Widen the icon-only collapse range for right-side navbar actions to `1200px–1590px` and target the last 4 `ms-auto` items, preventing label overflow at mid-widths.

## Changes
- [shared/styles/partials/_mobile.scss](shared/styles/partials/_mobile.scss): add `:not(.quarto-color-scheme-toggle)` to the `.quarto-navigation-tool` hide selectors.
- [shared/styles/partials/_navbar.scss](shared/styles/partials/_navbar.scss): extend media query upper bound to `1399px → 1590px` and switch selector to `ul.navbar-nav.ms-auto > li:nth-last-child(-n+4) .nav-link .menu-text`.

## Before Desktop
before instructor navbar

## After Desktop
after instructor navbar

## Before Mobile
before instructor mobile navbar

## After Mobile
after mobile instructor navbar

## Test plan
- [ ] Verify theme toggle is visible on mobile widths (<768px) on the instructors page.
- [ ] Verify right-side navbar actions collapse to icons cleanly between 1200px and 1590px.
- [ ] Confirm no regression at ≥1600px (full labels) and ≥hamburger breakpoint (mobile menu).
- [ ] Toggle dark/light mode from mobile to confirm the switch still works.

### DIFF
--- a/shared/styles/partials/_mobile.scss
+++ b/shared/styles/partials/_mobile.scss
@@ -8,8 +8,8 @@
   // UX: Hide ALL sidebar elements - we'll rely on hamburger menu for everything
   .sidebar-toggle,
   .sidebar-tools,
-  .quarto-navigation-tool,
-  .quarto-navigation .quarto-navigation-tool {
+  .quarto-navigation-tool:not(.quarto-color-scheme-toggle),
+  .quarto-navigation .quarto-navigation-tool:not(.quarto-color-scheme-toggle) {
     display: none !important;
   }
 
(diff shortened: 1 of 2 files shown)


profvjreddi:
Hey @farhan523 — thanks for digging into this! The mobile theme-toggle visibility fix in `_mobile.scss` is exactly right and is now merged in via this branch.

For the `_navbar.scss` change: that breakpoint range was independently fixed on `dev` in e1e2ba783 ("fix(navbar): keep right-side tools on one row across 1200-1479 px"), based on Playwright width measurements (brand 340 px + six dropdowns 680 px + right tools 445 px ≈ 1465 px needed → 1479 px upper bound, plus tightened 0.45 rem padding so icons fit at 1200 px). I resolved the conflict by taking `dev`'s version of `_navbar.scss` so the PR cleanly carries just your `_mobile.scss` fix.

Credit where it's due though — your investigation flagged the same regression family.

@all-contributors please add @farhan523 for bug, code

PR #1606, opened 2026-04-30
fix(kits/raspi/llm): wrap context-token array to prevent page overflow

farhan523:
## Summary
- On `kits/contents/raspi/llm/llm.qmd`, the 26-token `context` array was authored as **inline code** (single backtick) under a sub-bullet. Inline code does not wrap, so the array stretched the page wider than the viewport and forced horizontal scroll on the entire page (see *Before*).
- Moved the array into a fenced ```` ```text ```` code block with the values broken across three lines. Fenced code blocks scroll within their own width, so the page itself no longer overflows.
- Content of the array is unchanged; only formatting.

## Before
Screenshot 2026-04-30 101044

## After
Screenshot 2026-04-30 101116

## Test plan
- [ ] `cd kits && quarto preview` → open `/contents/raspi/llm/llm.html`.
- [ ] Scroll to the section after "*As we can see, several pieces of information are generated*".
- [ ] Confirm the page itself has no horizontal scrollbar at standard widths (e.g. 1280px, 1024px).
- [ ] Confirm the token array is rendered inside a code block and scrolls within its own bounds if the viewport is narrowed.
- [ ] Verify other code blocks on the page still render correctly.

### DIFF
--- a/kits/contents/raspi/llm/llm.qmd
+++ b/kits/contents/raspi/llm/llm.qmd
@@ -564,7 +564,12 @@ As we can see, several pieces of information are generated, such as:
     - `The capital of France is **Paris**. 🇫🇷`
 
 - **context**: the token IDs representing the input and context used by the model. Tokens are numerical representations of text used for processing by the language model.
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thank you so much for including the screenshot. It really helps me merge things in faster @farhan523!

profvjreddi:
yeah this was overflowing the page width in the PDF build. thanks @farhan523!

@all-contributors please add @farhan523 for bug in Kits

PR #1608, opened 2026-04-30
fix(mlsysim): skip viz test when matplotlib is not installed

Shashank-Tripathi-07:
## Summary

`test_scorecard_plot_accepts_scenario_evaluation_quantities` called `plot_evaluation_scorecard` unconditionally, but `matplotlib` is an optional dependency under `mlsysim[viz]` -- not installed in the base dev environment. Running `pytest` without the viz extra caused a hard `ImportError` instead of a graceful skip:

```
mlsysim/viz/plots.py:34: ImportError: matplotlib is required for plot generation.
FAILED tests/test_evaluation_contract.py::test_scorecard_plot_accepts_scenario_evaluation_quantities
```

**Fix:** Add `pytest.importorskip("matplotlib")` at the start of the test, which is the standard pytest pattern for optional-dependency tests -- the test is skipped when matplotlib is absent and runs normally when `mlsysim[viz]` is installed.

## Before / after

```
# Before (base dev env, no matplotlib):
FAILED test_evaluation_contract.py::test_scorecard_plot_accepts_scenario_evaluation_quantities

# After:
385 passed, 3 skipped in 1.92s
```

## Test plan

- [ ] `uv run python -m pytest tests/` with base deps -- confirm test is skipped, not failed
- [ ] `uv run python -m pytest tests/ --extras viz` (or with matplotlib installed) -- confirm test runs and passes


profvjreddi:
Thanks Rocky!

@all-contributors please add @Shashank-Tripathi-07 for bug in MLSysIM

PR #1623, opened 2026-05-02
fix(kits): improve dark mode dropdown menu contrast

farhan523:
## Summary
- fix Kits dark mode navbar dropdown background so menus no longer render on a white surface
- improve dropdown item, header, divider, and hover-state contrast for readability

## Before
- 
Screenshot 2026-05-02 090634

## After
- 
Screenshot 2026-05-02 090626

## Testing
- verified in local Kits preview in dark mode

### DIFF
--- a/kits/assets/styles/dark-mode.scss
+++ b/kits/assets/styles/dark-mode.scss
@@ -348,6 +348,38 @@ div[class*="margin"] {
   }
 }
 
+// Navbar dropdown menus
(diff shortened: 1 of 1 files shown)


profvjreddi (approved):
thanks @farhan523! confirmed there's no dropdown styling in the kits dark-mode file and the shared mobile.scss only handles the hamburger, so this is a real gap. lgtm

PR #1627, opened 2026-05-03
fix(labs): use ESP32 SRAM capacity and correct OOM badge in lab 03

Shashank-Tripathi-07:
## Summary

- `Hardware.Tiny.ESP32_S3.memory.capacity` returns 8 MB **flash storage** (7812 KiB), not SRAM
- The correct field for the 512 KiB SRAM is `memory.sram_capacity`
- Lab 03 line 39 was reading flash instead of SRAM, making `ESP32_RAM_KB = 7812` instead of `512`
- This caused the dynamic OOM ratio on line 348 (`~{size}x over`) to render as `~6x` instead of the correct `~98x`
- The hardcoded badge `200x OOM Discovered at Stage 5` is updated to `~100x OOM Discovered at Stage 5` to match

Same root cause as PR #1625 (lab 01) and PR #1626 (lab 02).

## Changes

```python
# Before (wrong -- reads 8 MB flash storage)
ESP32_RAM_KB = Hardware.Tiny.ESP32_S3.memory.capacity.m_as("KiB")

# After (correct -- reads 512 KiB SRAM)
ESP32_RAM_KB = Hardware.Tiny.ESP32_S3.memory.sram_capacity.m_as("KiB")
```

Badge: `200x OOM Discovered at Stage 5` -> `~100x OOM Discovered at Stage 5`

## Test plan

- [ ] Verify `Hardware.Tiny.ESP32_S3.memory.sram_capacity.m_as("KiB")` returns 512
- [ ] Run lab 03 and confirm the Engine.solve() block shows `~98x over`
- [ ] Confirm badge reads `~100x OOM`

PR #1626, opened 2026-05-03
fix(labs): use ESP32 SRAM capacity field in lab 02

Shashank-Tripathi-07:
## Summary

- `Hardware.Tiny.ESP32_S3.memory.capacity` returns 8 MB **flash storage** (7812 KiB), not SRAM
- The correct field for the 512 KiB SRAM is `memory.sram_capacity`
- Lab 02 line 58 was reading flash instead of SRAM, making `ESP32_RAM_KB = 7812` instead of `512`

Same root cause as PR #1625 (lab 01 fix).

## Change

```python
# Before (wrong -- reads 8 MB flash storage)
ESP32_RAM_KB = Hardware.Tiny.ESP32_S3.memory.capacity.m_as("KiB")

# After (correct -- reads 512 KiB SRAM)
ESP32_RAM_KB = Hardware.Tiny.ESP32_S3.memory.sram_capacity.m_as("KiB")
```

## Test plan

- [ ] Verify `Hardware.Tiny.ESP32_S3.memory.sram_capacity.m_as("KiB")` returns 512 in the mlsysim registry
- [ ] Run lab 02 in Marimo and confirm `ESP32_RAM_KB` shows 512 KiB

PR #1625, opened 2026-05-03
fix(labs): correct ESP32 SRAM field and OOM ratio in Lab 01

Shashank-Tripathi-07:
## Bug

`labs/vol1/lab_01_ml_intro.py` line 67 reads `ESP32.memory.capacity` to get the ESP32's memory size, but `memory.capacity` returns the **8 MB flash storage**, not SRAM.

The correct field is `ESP32.memory.sram_capacity` = **512 KiB**.

## Impact

`ESP32_RAM_KB` was 7812 (flash) instead of 512 (SRAM), making the OOM ratio compute as ~6x instead of the correct ~98x. Every downstream callout, badge, and answer reveal in Part C showed fabricated numbers.

## Fix

- Line 67: `ESP32.memory.capacity.m_as("KiB")` → `ESP32.memory.sram_capacity.m_as("KiB")`
- Badge: `200x Out of Memory` → `~100x Out of Memory`
- Learning objective: "infeasible by 200x" → "infeasible by ~100x"
- Part C question option C: `~200x over budget` → `~100x over budget`
- Answer reveal text: updated hardcoded "200x" references to match actual ~98x ratio

## Verification

```python
from mlsysim import Hardware, Models
ESP32 = Hardware.Tiny.ESP32_S3
RESNET50_PARAMS = Models.ResNet50.parameters.m_as('count')
RESNET50_SIZE_MB = RESNET50_PARAMS * 2 / (1024*1024)  # FP16

# Before fix (wrong)
wrong_kb = ESP32.memory.capacity.m_as('KiB')        # 7812 KiB (flash)
print(RESNET50_SIZE_MB * 1024 / wrong_kb)            # ~6x -- wrong

# After fix (correct)
correct_kb = ESP32.memory.sram_capacity.m_as('KiB')  # 512 KiB (SRAM)
print(RESNET50_SIZE_MB * 1024 / correct_kb)          # ~98x -- correct
```

PR #1628, opened 2026-05-03
fix(labs): use ESP32 SRAM capacity field in lab 07

Shashank-Tripathi-07:
## Summary

`ESP32.memory.capacity` returns 8 MB **flash storage** (7812 KiB), not SRAM. The correct field is `memory.sram_capacity` (512 KiB).

In lab 07, `ESP32_RAM_KB` is used dynamically in:
- The deployment target label: `ESP32 ({ESP32_RAM_KB:.0f} KB)` -- showed `7812 KB` instead of `512 KB`
- The MathPeek block: `ESP32 memory: {ESP32_RAM_KB:.0f} KB = {ESP32_RAM_KB/1024:.2f} MB` and `Ratio: 1800 / {ESP32_RAM_KB/1024:.2f} = {ratio:,.0f}x` -- showed `7.63 MB` and `~236x` instead of `0.50 MB` and `~3,600x`

The hardcoded explanation text throughout the lab correctly states "3,600x" and "512 KB", so the wrong variable caused a visible contradiction.

## Change

```python
# Before (wrong -- reads 8 MB flash storage)
ESP32_RAM_KB = ESP32.memory.capacity.m_as("KiB")

# After (correct -- reads 512 KiB SRAM)
ESP32_RAM_KB = ESP32.memory.sram_capacity.m_as("KiB")
```

Same root cause as PRs #1625 (lab 01), #1626 (lab 02), #1627 (lab 03).

## Test plan

- [ ] Run lab 07 Part D, confirm deployment target label shows `ESP32 (512 KB)`
- [ ] Confirm MathPeek shows `0.50 MB` and `~3,600x`

### DIFF
--- a/labs/vol1/lab_07_ml_frameworks.py
+++ b/labs/vol1/lab_07_ml_frameworks.py
@@ -40,7 +40,7 @@ async def _():
     H100_BW_GBS = H100.memory.bandwidth.m_as("GB/s")
     H100_DISPATCH = H100.dispatch_tax.m_as("ms")
     H100_RAM_GB = H100.memory.capacity.m_as("GB")
-    ESP32_RAM_KB = ESP32.memory.capacity.m_as("KiB")
+    ESP32_RAM_KB = ESP32.memory.sram_capacity.m_as("KiB")
 
     # Framework runtime overheads (MB) -- from textbook Table 7.x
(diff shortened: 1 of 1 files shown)


profvjreddi:

PR #1630, opened 2026-05-03
fix(labs): correct FLOP scaling answer key in lab 05 Part C

Shashank-Tripathi-07:
## Summary

Part C asks: "A 3-layer MLP has hidden layers of width 128. You double hidden width to 256. By how much do total FLOPs increase?"

The answer key marked **"C) ~4x (quadratic)"** as correct, but the code itself computes:

```
_actual_256 = (2*784*256 + 2*256*256 + 2*256*10) / (2*784*128 + 2*128*128 + 2*128*10)
            = 537,600 / 236,032
            ≈ 2.28x
```

The 4x scaling only applies to the hidden-to-hidden layer (2*W^2). The input layer (2*784*W) and output layer (2*W*10) scale linearly, pulling the total to ~2.3x. The correct answer is **"A) 2x"**.

The "Correct" callout itself was already printing `~{_actual_256:.1f}x` = `~2.3x` while saying "Correct" for 4x -- a visible contradiction. The MathPeek also said "(not 2x!)" when it should say "(not 4x!)".

## Changes

- Correct answer: `"4x"` → `"2x"`
- "2x" callout: now shows "Correct" with accurate explanation
- "4x" callout: now shows why it's wrong (quadratic layer alone is 4x, but total is ~2.3x)
- MathPeek note: `(not 2x!)` → `(not 4x!)`

## Test plan

- [ ] Open lab 05 Part C, confirm the live ratio displays ~2.3x
- [ ] Select "A) 2x" -- confirm "Correct" callout appears
- [ ] Select "C) ~4x" -- confirm correction callout explains the ~2.3x actual value

### DIFF
--- a/labs/vol1/lab_05_nn_compute.py
+++ b/labs/vol1/lab_05_nn_compute.py
@@ -831,16 +831,18 @@ The quadratic term dominates as W grows.
 
         _actual_256 = (2*784*256 + 2*256*256 + 2*256*10) / (2*784*128 + 2*128*128 + 2*128*10)
         _pred = partC_prediction.value
-        if _pred == "4x":
+        if _pred == "2x":
             items.append(mo.callout(mo.md(
                 f"**Correct.** Doubling width from 128 to 256 increases total FLOPs by "
-                f"~{_actual_256:.1f}x. The hidden-to-hidden layer scales as W^2. "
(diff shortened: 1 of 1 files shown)


profvjreddi:

PR #1682, opened 2026-05-05
fix(slides): readable section-list tables on tinyml pages 

farhan523:
## Summary
- `.module-card` hardcoded `background: #ffffff` while dark-mode.scss colored all `<td>` light gray, making row text invisible on `tinyml-fundamentals.html` (and the other `tinyml-*` pages that share the component).
- Adds dark-mode overrides for `.module-card` (background, headings, description, companion pill), `.section-list` (th/td color, borders, hover, sec-title, no-material), and `.section-header-sm`.
- Tunes the `.badge-slides`, `.badge-reading`, and `.badge-colab` swatches so the Slides/Reading/Colab pills keep adequate contrast on the dark card.

Before:
Screenshot 2026-05-05 195721

After:
Screenshot 2026-05-05 195736

## Test plan
- [ ] Open `tinyml-fundamentals.html` in dark mode and confirm every row of every section-list is readable.
- [ ] Spot-check `tinyml-applications.html`, `tinyml-deploying.html`, `tinyml-mlops.html` for the same components.
- [ ] Toggle back to light mode and verify nothing regressed (cards still white, badges keep their original colors).

### DIFF
--- a/slides/assets/styles/dark-mode.scss
+++ b/slides/assets/styles/dark-mode.scss
@@ -207,6 +207,57 @@ pre code {
   color: $slides-accent-dark !important;
 }
 
+// TinyML course pages: module cards + section lists
+.module-card {
+  background: $bg-elevated !important;
+  border-color: $border-color-dark !important;
+
+  h3, h4 { color: $body-color-dark !important; }
+
+  .module-desc { color: $text-muted-dark !important; }
+
+  .companion {
+    background: rgba(255, 255, 255, 0.04) !important;
+    color: $text-muted-dark !important;
+  }
+  .companion a { color: $slides-accent-dark !important; }
+}
+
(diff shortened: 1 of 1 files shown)


profvjreddi:
Merged this. The fix is in the right layer: these TinyML module-card / section-list styles are slide-site components, not shared ecosystem primitives, so keeping the dark-mode overrides in `slides/assets/styles/dark-mode.scss` is appropriate.

I also added a small maintainer follow-up on `dev` after merge: `08976875962c` scopes the dark-mode badge color overrides under `.section-list` so the `.badge-slides`, `.badge-reading`, and `.badge-colab` rules stay tied to the TinyML table UI instead of becoming broader global badge overrides. That keeps the immediate contrast fix while reducing future maintenance risk.

PR #1686, opened 2026-05-06
fix(tinytorch): strip solution blocks when generating student notebook via tito module start

### LINKED ISSUE #1684: [TinyTorch] Answers already in module 1
### Module

01 Tensor

### Type of Improvement

Code quality (refactor, style, consistency)

### Description

The solutions were already implemented when I started module 1.
```
# Install TinyTorch (run from a project folder like ~/projects)
curl -sSL mlsysbook.ai/tinytorch/install.sh | bash

# Activate and verify
cd tinytorch
source .venv/bin/activate
tito setup

tito module start 01
```

### Proposed Solution

Remove the implementations from the working notebooks.

Shashank-Tripathi-07:
## Bug

Closes #1684

`tito module start 01` was passing the full `src/01_tensor/01_tensor.py` -- including all `### BEGIN SOLUTION` / `### END SOLUTION` blocks -- directly to jupytext. Students received notebooks with every answer already implemented, defeating the purpose of the exercises.

The docstring in `_create_module_from_src` even acknowledged this explicitly:

> "Full `src/` (including `### BEGIN SOLUTION` ... `### END SOLUTION` blocks) is passed through to jupytext so notebooks match the source-of-truth"

## Root cause

`_create_module_from_src` in `workflow.py` called `convert_py_to_notebook` on the raw source path with no stripping step. Solution stripping was not implemented anywhere in the `tito module` path.

## Fix

Before passing to jupytext, strip solution blocks in memory and write to a temp directory:

```python
lines = src_file.read_text(encoding="utf-8").splitlines(keepends=True)
stripped = []
in_solution = False
for line in lines:
    if line.strip() == "### BEGIN SOLUTION":
        in_solution = True
        continue
    if line.strip() == "### END SOLUTION":
        in_solution = False
        continue
    if not in_solution:
        stripped.append(line)
```

The stripped source is written to a `tempfile.TemporaryDirectory`, jupytext converts it, and the temp dir is cleaned up automatically. The original `src/` files are never modified.

## Verification

After the fix, `tito module start 01` generates `modules/01_tensor/tensor.ipynb` with empty `pass` stubs and TODO comments -- no implemented solutions.

### DIFF
--- a/tinytorch/tito/commands/module/workflow.py
+++ b/tinytorch/tito/commands/module/workflow.py
@@ -390,22 +390,52 @@ class ModuleWorkflowCommand(BaseCommand):
         return self._open_jupyter(module_name)
 
     def _create_module_from_src(self, module_name: str) -> bool:
-        """Create a module in modules/ by converting from src/.
+        """Create a student-facing module notebook in modules/ from src/.
 
-        Uses the same conversion logic as 'tito src export' but only creates
-        the student-facing notebook, without exporting to the tinytorch package.
-        Full `src/` (including `### BEGIN SOLUTION` ... `### END SOLUTION` blocks) is
-        passed through to jupytext so notebooks match the source-of-truth and exports
-        remain consistent for `tito module complete` and CI user-journey.
+        Strips ### BEGIN SOLUTION ... ### END SOLUTION blocks so students
+        receive blank stubs, not implemented answers.
         """
+        import tempfile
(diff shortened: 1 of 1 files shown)


profvjreddi:
Closing — the solution-stripping logic will be revisited in a future pass when the TinyTorch module workflow is updated. Thanks for the contribution.

PR #1690, opened 2026-05-06
docs(tinytorch): add GELU implementation and explanation

### LINKED ISSUE #1689: [TinyTorch] GELU implementation missing from module 2 docs
### Module

02 Activations

### Type of Improvement

Better explanation or documentation

### Description

The GELU implementation is missing for the [core concepts](https://mlsysbook.ai/tinytorch/modules/02_activations.html#core-concepts) section.

### Proposed Solution

Include this code as implementation along with necessary GELU supporting info. Logically this section seems to belong after `#sigmoid-and-tanh` and before `#softmax-and-numerical-stability`.

```py
class GELU:
    def forward(self, x: Tensor) -> Tensor:
        """Apply GELU activation using the fast sigmoid approximation."""
        return Sigmoid()(x * 1.702) * x
```

bdub-1:
## Problem
Closes #1689 

The "Activations" module (Module 02) mentioned the GELU activation function is referenced throughout the docs and is required in the coding module, but the actual implementation and explanation were missing from the core content.

## Fix
- Added the `GELU` class implementation to `02_activations.qmd` 
- Documented the computational tradeoff (4–5x cost vs ReLU)

### DIFF
--- a/tinytorch/quarto/modules/02_activations.qmd
+++ b/tinytorch/quarto/modules/02_activations.qmd
@@ -552,10 +552,7 @@ Below is the complete implementation from your module:
 class GELU:
     def forward(self, x: Tensor) -> Tensor:
         """Apply GELU activation using the fast sigmoid approximation."""
-        # GELU approximation: x * sigmoid(1.702 * x)
-        sigmoid_part = 1.0 / (1.0 + np.exp(-1.702 * x.data))
-        result = x.data * sigmoid_part
-        return Tensor(result)
+        return Sigmoid()(x * 1.702) * x
 ```
 
 While ReLU remains the default in CNNs and MLPs, GELU has become the standard choice for modern transformers like GPT and BERT.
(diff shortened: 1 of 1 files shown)


profvjreddi:
Reviewed this against the TinyTorch module/source layout before merging.

This belongs in `tinytorch/quarto/modules/02_activations.qmd`: that file is authored website/module prose, not the generated notebook release path that made #1686 the wrong approach. I pushed a small cleanup on top before merge to keep it maintainable:

- removed trailing whitespace in the `### GELU` heading/prose
- changed the wording from a loose “complete implementation” claim to the page’s normal implementation-description style
- added a manual GELU listing caption
- bumped the following Softmax listing number from 2.2 to 2.3 so the section stays internally consistent

Verification:

- `codespell tinytorch/quarto/modules/02_activations.qmd`
- `quarto render modules/02_activations.qmd --to html`

The targeted render succeeds. It still reports the two pre-existing unresolved listing-reference warnings for Sigmoid and Softmax on single-page render; this change does not add a new one.

PR #1821, opened 2026-05-29
fix(staffml): give ChainStrip dots accessible names and a current-step marker

farhan523:
## Summary
The progress dots in `ChainStrip` conveyed state purely by color, scale, and opacity (`scale-125` for current, `opacity-60` for past, `opacity-30` for future), and the only accessible name on each `<button>` was a `title=` — which screen readers announce inconsistently. Keyboard and screen-reader users couldn't tell which step was current or what each step was about.

### Changes
- Wrap the dots in `<ol aria-label="Question chain progress">` with each dot in an `<li>` so AT announces this as an ordered list of N items rather than a loose bag of buttons.
- Mark the current dot with `aria-current="step"`.
- Give each dot an `aria-label` of the form `Part X of Y: <level> <name> — <title>` (with `" (completed)"` appended for past steps), so the accessible name doesn't depend on `title=`.
- Add `type="button"` to the dots so they can't accidentally submit a form if the strip is ever rendered inside one.

### Visual impact
Zero. The list resets default `list-style` / padding / margin so the dot row still renders identically.

## Test plan
- [x] `npx vitest run` → 43/43 passing (was 37/37 on dev — 6 new tests for list semantics, single `aria-current`, accessible-name format, completed-suffix on past steps only, `type="button"` on every dot, and that `onNavigate` is still wired to clicks)
- [x] `npx tsc --noEmit` → clean
- [ ] Manual a11y: open a chained question, Tab into the dot row, listen — screen reader should announce the list, then each dot with its part/level/title, with the current dot identified as the current step
- [ ] Manual visual: dot row should look identical to before

### DIFF
--- a/interviews/staffml/src/__tests__/chain-strip.test.tsx
+++ b/interviews/staffml/src/__tests__/chain-strip.test.tsx
@@ -1,0 +1,82 @@
+/**
+ * ChainStrip a11y guardrails.
+ *
+ * The progress dots used to convey state purely by color/scale/opacity
+ * and `title=` attributes — keyboard / screen-reader users couldn't tell
+ * which step was current or what each step was about. We now render the
+ * dots in an `<ol>` with `aria-current="step"` on the active dot and an
+ * accessible name on each button.
+ */
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen, within } from '@testing-library/react';
+import ChainStrip from '@/components/ChainStrip';
+import type { ChainInfo } from '@/lib/corpus';
+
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07 (approved):
Very good bro of making the interface accessible for screen reader apps also 👏🏻

farhan523:
> Very good bro of making the interface accessible for screen reader apps also 👏🏻

Hi @Shashank-Tripathi-07 thank you bro for the appreciation, I think we should try to make this content accessible for everyone.

PR #1834, opened 2026-06-03
fix(staffml): pause Nav due-count poll on hidden tabs + sync across tabs

farhan523:
## Summary
Nav was polling `getDueCount` every 30s indefinitely via a bare `setInterval`, even on backgrounded tabs. Cheap per call (it reads localStorage), but the unnecessary wakeups burn battery on mobile and add no value when no one is looking. There was also no cross-tab sync: a card completed in one tab wouldn't update the badge in another tab until that other tab's next 30s tick.

### Changes
- **New hook** `src/lib/hooks/useVisibilityPoll.ts`: pauses the interval on `visibilitychange→hidden`, re-arms on `→visible` with an immediate catch-up tick. Callback is held in a ref so a fresh closure on every render is picked up without re-arming.
- **Throw-isolation inside the hook**: each tick is wrapped in `try/catch` with `console.error`. Without this, a throwing callback on the mount-time synchronous tick propagates out of the effect body *before* React receives the cleanup — the already-armed `setInterval` would orphan and re-throw every interval forever.
- **Cross-tab freshness in Nav**: the hook is explicit in its docs that it is **not** a cross-tab sync primitive (only catches up on visibility transitions, not on two-tabs-both-visible). So Nav adds a separate `storage` event listener that fires in *other* tabs the instant one writes to localStorage.
- **Replace `Nav`'s empty `catch {}` with `console.error(...)`** so a real failure (corrupted localStorage, JSON parse error) surfaces in devtools instead of presenting as a frozen badge. Refresh logic hoisted into one `useCallback` shared by both the poll and the storage handler.

## Test plan
- [x] `npx vitest run` → 65/65 individual cases (was 57 on dev; +8 new for the hook).
- [x] `npx tsc --noEmit` → unchanged from dev baseline (the 2 pre-existing `katex` / `js-quantities` errors are env-local: those packages are declared in `package.json` but missing from `node_modules` until `npm install` runs).
- [ ] Manual: open Nav in two browser windows, complete an SR card in window A, observe the due-count badge in window B drop within ~1 tick (the `storage` event fires immediately; the visibility poll would otherwise wait up to 30s).
- [ ] Manual: open DevTools → Application → throttle background tab to confirm no `setInterval` wakeups while the tab is hidden.

### DIFF
--- a/interviews/staffml/src/__tests__/use-visibility-poll.test.tsx
+++ b/interviews/staffml/src/__tests__/use-visibility-poll.test.tsx
@@ -1,0 +1,144 @@
+/**
+ * Tests for useVisibilityPoll — the Page Visibility-aware setInterval
+ * wrapper extracted from Nav's SR-due-count poll.
+ *
+ * Pins the four behaviors that make this hook useful versus a plain
+ * setInterval:
+ *   1. Fires immediately on mount (so the UI isn't blank for one interval).
+ *   2. Fires repeatedly at the requested cadence while visible.
+ *   3. Stops firing when the tab goes hidden.
+ *   4. Catches up with an immediate call when the tab comes back, then
+ *      resumes the cadence.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook } from '@testing-library/react';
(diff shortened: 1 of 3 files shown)


profvjreddi:

PR #1843, opened 2026-06-06
fix(staffml): derive PaperCitationCard year from buildDate, not `new Date()`

farhan523:
## Summary
The BibTeX entry in `PaperCitationCard` took its year from `new Date().getFullYear()` — the *render-time* year on the client. With Next's static export the HTML is generated at build time and hydrates on the client whenever the user visits, so server (build) and client (hydrate) could disagree on the year:

- around midnight UTC on Dec 31 / Jan 1, OR
- any time the site is viewed in a year after the build year (i.e. always, eventually).

Hydration mismatch warning either way, and the wrong anchor for a citation regardless — the year in a software citation should be the release year, not the year the reader happened to view the page.

### Changes
- **Make `buildDate` required** on `PaperCitationCardProps`. The only call site at [`app/about/page.tsx`](src/app/about/page.tsx#L70) already passes it from `BUILD_DATE` in `lib/stats`, so this is a type-level alignment with the existing real-world usage.
- **Derive the BibTeX year** via `new Date(buildDate).getUTCFullYear()` so the year matches the build server's UTC reference regardless of viewer timezone.
- `buildDateLabel` simplifies: no `buildDate ? ... : null` ternary, just the formatted string. The `&&` short-circuit at the render site is now redundant but harmless — left as defense-in-depth.

### Out of scope
The displayed `buildDateLabel` still uses `toLocaleDateString` with no timezone option, which CAN mismatch on a build like `2026-04-22T03:00:00Z` viewed in PT (renders \"April 22, 2026\" on the build server, \"April 21, 2026\" on a PT client). That's a separate, smaller hydration concern with a UX trade-off (UTC vs viewer's local date) — happy to file a follow-up if it's worth fixing.

## Test plan
- [x] `npx vitest run` → 60/60 individual cases (was 57 on dev; +3 new for the year derivation, including a clock-mock test that catches future regressions to `new Date()`).
- [x] `npx tsc --noEmit` → unchanged from dev baseline (the 2 pre-existing `katex` / `js-quantities` errors are env-local).
- [ ] Manual: open `/about`, expand \"Cite this release\", confirm the BibTeX year matches `BUILD_DATE`'s UTC year.

### DIFF
--- a/interviews/staffml/src/__tests__/paper-citation-card.test.tsx
+++ b/interviews/staffml/src/__tests__/paper-citation-card.test.tsx
@@ -1,0 +1,66 @@
+/**
+ * PaperCitationCard year stability.
+ *
+ * Previously the BibTeX year came from `new Date().getFullYear()`, which
+ * is the *render-time* year on the client. With Next's static export the
+ * HTML is generated at build time and hydrates on the client whenever the
+ * user visits — server (build) and client (hydrate) could disagree on the
+ * year (e.g. around midnight UTC on Dec 31 / Jan 1, or just any time the
+ * site is viewed in a year after the build year). That's a hydration
+ * mismatch AND, more fundamentally, the wrong anchor for a citation.
+ * The year now comes from the required `buildDate` prop (UTC).
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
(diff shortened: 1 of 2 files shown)


profvjreddi:

PR #1845, opened 2026-06-06
fix(staffml): trap focus inside the AskInterviewer waitlist modal

farhan523:
## Summary
The waitlist modal at the bottom of `AskInterviewer` carried `role=\"dialog\" aria-modal=\"true\"` but didn't enforce the contract — Tab and Shift+Tab escaped into background controls behind the backdrop, and the email input wasn't auto-focused on open. Keyboard users landed in the page body and had to hunt for the input.

### Changes
- **Auto-focus the email input on open** via the same `setTimeout(0)` pattern used by `CommandPalette` and `KeyboardShortcutsOverlay`. Keyboard users land in the right place.
- **Focus trap**: `Tab` from the last focusable cycles back to the first, `Shift+Tab` from the first cycles to the last. Middle-of-trap presses fall through to the browser's native handling — we only block at the edges.
- **Focus restoration**: capture `document.activeElement` on mount, restore it on unmount so focus returns to the trigger that opened the modal.
- **Export `WaitlistModal`** as a named export so the regression test can render it directly without spinning up the full `AskInterviewer` parent and triggering its open state.

Mirrors the focus-trap pattern already used in `CommandPalette` (lines 238-252), so the two modals behave identically for keyboard users.

## Test plan
- [x] `npx vitest run` → 62/62 individual cases (was 57 on dev; +5 new for autofocus, restoration, forward cycle, backward cycle, middle-trap fall-through).
- [x] `npx tsc --noEmit` → unchanged from dev baseline (the 2 pre-existing `katex` / `js-quantities` errors are env-local).
- [ ] Manual: open Mock Interview, click \"Join waitlist\" (or whatever triggers `WaitlistModal`), confirm the email input has focus, Tab repeatedly and verify focus stays inside the dialog, close and verify focus returns to the trigger.

### DIFF
--- a/interviews/staffml/src/__tests__/waitlist-modal.test.tsx
+++ b/interviews/staffml/src/__tests__/waitlist-modal.test.tsx
@@ -1,0 +1,96 @@
+/**
+ * WaitlistModal a11y: focus management + focus trap.
+ *
+ * Before: the dialog had role="dialog" aria-modal="true" but Tab/Shift+Tab
+ * escaped the modal into background controls behind the backdrop, and the
+ * email input wasn't auto-focused on open — keyboard users landed in the
+ * page body and had to hunt for the input.
+ *
+ * After: focus moves to the email input on open, Tab cycles within the
+ * surface, and focus returns to the previously-focused element on unmount.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks for the fix. Sorry, I'm a little bit behind. I've been very busy making lots of updates to code etc. Will look now.

profvjreddi:

PR #1854, opened 2026-06-10
fixes for tinytorch modules 07 and 08 as well as some rewording for lab 01

vedant-a-joshi:
Tested through `tito module test NN` for modules 07 and 08.
_Note_: It fails FAILED tests/integration/test_training_flow.py::TestTrainingReducesLoss::test_mlp_loss_decreases, however the commits here aren't meant to fix this, and they shouldn't be altering/causing it in any way.

I manually checked to make sure the new diagrams and wording is accurate.

Changes for tinytorch
* Fixed diagram representations for easier understanding
* Fixed alignment/consistency issues

Changes for labs
* Fixed duration so that it adds up in the correct spot
* Fixed equation representation
* Fixed graph so that logarithm y-axis so that y-axis is easier to understand

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2,7 +2,7 @@
     "projectName": "cs249r_book",
     "projectOwner": "harvard-edge",
     "files": [
-        "book/quarto/contents/frontmatter/acknowledgements/acknowledgements.qmd",
+        "/contents/frontmatter/acknowledgements/acknowledgements.qmd",
         "README.md"
     ],
     "contributors": [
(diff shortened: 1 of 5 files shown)


profvjreddi:
Appreciate this, @vedant-a-joshi, and your catch was real: the old lab's duration and the 200x ratio were inconsistent.

The trouble is `lab_01_ml_intro.py` was rewritten on dev into the new track-aware D-A-M lab since you branched, so those exact values are gone (the new lab is structured as four parts plus a memo). That makes the lab hunks here impossible to merge directly, and the PR also rolls back two generated files (`.all-contributorsrc` to a path that does not exist, and `contributors.json` to older stats).

If you can rebase on current dev and reapply any duration or ratio fixes that still apply to the new lab, I will gladly take them. The tinytorch 07/08 diagram tweaks would be easiest as their own small PR. Thanks for the careful eye.

profvjreddi:
@all-contributors please add @vedant-a-joshi for code, doc

profvjreddi:
Closing this out. The tinytorch module 07 and 08 fixes are integrated into dev with CI green, and you are credited (code, doc). Thanks again @vedant-a-joshi.

PR #1870, opened 2026-06-16
fix(activations): numerically stable sigmoid + remove overflow warning suppressor

Shashank-Tripathi-07:
## What breaks

`Sigmoid.forward()` uses the naive form:
```python
result = 1.0 / (1.0 + np.exp(-x.data))
```

For very negative inputs (`x << -87` in float32), `np.exp(-x)` overflows to `inf`. The result `1/(1+inf) = 0.0` is accidentally correct, but NumPy emits a `RuntimeWarning`. The unit test hid this with:
```python
warnings.filterwarnings('ignore', category=RuntimeWarning)
```

This suppressor silences the whole `RuntimeWarning` category for any future code in that block -- meaning a real numerical regression would produce no signal.

## Fix

Use the piece-wise stable sigmoid identity:
```python
result = np.where(
    x_data >= 0,
    1.0 / (1.0 + np.exp(-x_data)),    # exp(-x) in [0,1], safe
    np.exp(x_data) / (1.0 + np.exp(x_data))  # exp(x) in [0,1), safe
)
```

Both branches compute the same mathematical function but stay in the safe exponent range. This is the same trick used by `scipy.special.expit` and PyTorch's `torch.sigmoid`.

With overflow eliminated: remove the `warnings.filterwarnings` calls and the now-unused `import warnings`.

## Test plan
- [ ] `pytest tinytorch/tests/02_activations/` passes with no RuntimeWarning
- [ ] `sigmoid([-1000, 1000])` returns `[~0, ~1]` without any warning
- [ ] `sigmoid([0])` still returns `0.5`

### DIFF
--- a/tinytorch/src/02_activations/02_activations.py
+++ b/tinytorch/src/02_activations/02_activations.py
@@ -92,7 +92,6 @@ If you see import errors, ensure you've run `tito export` after completing Modul
 import numpy as np
 rng = np.random.default_rng(7)
 from typing import Optional
-import warnings
 
 # Import from TinyTorch package (previous modules must be completed and exported)
 from tinytorch.core.tensor import Tensor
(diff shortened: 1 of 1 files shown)


profvjreddi:
This is landed, @Shashank-Tripathi-07, thanks. You were right that the test was hiding a real overflow by suppressing the whole warning category.

I kept your piecewise stable form but wrapped it in `np.errstate`, since `np.where` evaluates both branches and the discarded one can still overflow. That makes it genuinely silent, so the test no longer needs the blanket suppressor. Pushed to dev as `ecee1841ea` with your authorship. Nice fix.

PR #1875, opened 2026-06-16
fix(attention): mask reshape assumes square (seq_q==seq_k), breaks cross-attention

Shashank-Tripathi-07:
## What breaks

`MultiHeadAttention.forward()` reshapes a 3-D attention mask `(batch, seq_q, seq_k)` to `(batch, 1, seq_q, seq_k)` for multi-head broadcasting. The old code:

```python
batch_size_mask, seq_len_mask, _ = mask.shape   # discards seq_k!
mask_reshaped = mask.reshape(batch_size_mask, 1, seq_len_mask, seq_len_mask)
```

The `_` discards `seq_k`. The reshape then uses `seq_len_mask` (which is `seq_q`) for both the query and key dimensions. For self-attention where `seq_q == seq_k`, this accidentally works. For cross-attention (e.g. encoder-decoder attention, or attention with a different context length), `seq_q != seq_k`, so the reshape either:
- Raises `ValueError` if `seq_q * seq_q != seq_q * seq_k` (total elements don't match)
- Silently reinterprets memory if they happen to share a common product

## Fix

```python
batch_size_mask, seq_q_mask, seq_k_mask = mask.shape
mask_reshaped = mask.reshape(batch_size_mask, 1, seq_q_mask, seq_k_mask)
```

Unpack both dimensions explicitly. Correct for square and rectangular masks.

## Test plan
- [ ] `pytest tinytorch/tests/12_attention/` passes
- [ ] Self-attention with square mask `(B, T, T)` still works
- [ ] Cross-attention with rectangular mask `(B, T_q, T_k)` where `T_q != T_k` now works


profvjreddi:
Thanks @Shashank-Tripathi-07. Verified the fix is correct: the old reshape raises ValueError on cross-attention (seq_q != seq_k) and only works when seq_q == seq_k, while the new code reshapes to (batch, 1, seq_q, seq_k) and broadcasts across heads correctly, with self-attention unchanged.

This branch was based on `main`, so it could not merge cleanly into `dev` (our integration branch). I reapplied your change on `dev` as `a30b36cd76` with your authorship preserved, so it is now landed. Closing as merged. Thank you for the catch.

profvjreddi:
Thanks @Shashank-Tripathi-07! 🎉 Your attention, quantization, sigmoid, and module-complete harness fixes all landed on dev (#1875, #1874, #1870, #1877, #1878).

@all-contributors please add @Shashank-Tripathi-07 for bug, code in tinytorch

PR #1877, opened 2026-06-16
fix(module): syntax errors in student notebook silently pass tito module complete

Shashank-Tripathi-07:
## Bug report

Reported by @Chufeng-Jiang in discussion #1827.

A student had a syntax error in their Module 01 notebook (missing closing parenthesis in an `isinstance` call). Running `tito module complete 01` reported all tests passing. The error only surfaced when Module 02 tests tried to import from the broken `tinytorch/core/tensor.py` that the export had written.

## Root cause

`_run_inline_unit_tests()` runs the instructor's source file at `src/01_tensor/01_tensor.py`, not the student's notebook at `modules/01_tensor/tensor.ipynb`. A syntax error in the student's notebook is invisible to this test phase -- the instructor source always runs clean. The broken notebook then gets exported into `tinytorch/core/tensor.py` without any parse check.

## Fix

Add `_check_notebook_syntax()`: reads the student's notebook, compiles every code cell with Python's `compile()`, and returns the first `SyntaxError` found with the cell number and line. Called in both `complete_module` and `_complete_module_quiet` immediately before the export step, so the student sees:

```
❌ Syntax error in your notebook -- fix it before exporting
   SyntaxError in tensor.ipynb cell 12, line 3: '(' was never closed
     if len(shape) == 1 and isinstance(shape[0], (tuple, list):
```

instead of a confusing failure in the next module.

## Test plan

- [ ] Student notebook with a syntax error: `tito module complete 01` reports the error with cell + line number and exits 1
- [ ] Valid notebook: behaviour unchanged, export proceeds normally
- [ ] `tito module complete --all` (quiet path) also blocked on syntax errors

Co-authored-by: Chufeng Jiang <cjiang@gradcenter.cuny.edu>

### DIFF
--- a/tinytorch/tito/commands/module/workflow.py
+++ b/tinytorch/tito/commands/module/workflow.py
@@ -595,6 +595,12 @@ class ModuleWorkflowCommand(BaseCommand):
             self.console.print("[bold cyan] Step 2/4: Exporting to TinyTorch Package[/bold cyan]")
             self.console.print()
 
+            syntax_error = self._check_notebook_syntax(module_name)
+            if syntax_error:
+                self.console.print(f"[red]   ❌ Syntax error in your notebook -- fix it before exporting[/red]")
+                self.console.print(f"      [dim red]{syntax_error}[/dim red]")
+                return 1
+
             export_result = self.export_module(module_name)
             if export_result != 0:
                 self.console.print(f"[red]   ❌ Export failed for {module_name}[/red]")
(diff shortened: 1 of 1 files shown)


Chufeng-Jiang:
This problem happened again on module 6 when I miss spelled the "requires_grad" while it passed successfully the complete module 06 test.

image
image

profvjreddi:
Landed, @Shashank-Tripathi-07. Your notebook-syntax check is on dev as part of `16b246fd84`, kept additive so the unit tests still run before export.

I folded it together with the integration-test discovery fix from #1878 (the two touch the same area), so I'm closing this as merged into that commit. Thanks for catching that syntax errors were sailing straight through to export.

PR #1878, opened 2026-06-16
fix module complete: integration tests never ran and empty notebooks silently passed

Shashank-Tripathi-07:
### Bug 1: integration tests never ran for any module

`_run_integration_tests` looked for `test_progressive_integration.py` in each module's test directory. That file does not exist anywhere. The actual files follow the pattern `test_NN_<name>_progressive.py` (e.g. `test_01_tensor_progressive.py`, `test_06_autograd_progressive.py`). Because the file was never found, step 3 in `tito module complete` always returned `passed=0 failed=0` silently, meaning no student has ever had their exported implementation validated by integration tests.

Fixed by adding `_find_integration_test_file` which looks for `test_NN_*_progressive.py` first (matching the module number prefix), then falls back to any `test_*_progressive.py` in the module directory.

### Bug 2: empty notebooks silently passed unit tests

`_parse_test_output` entered a branch where `tests` stayed `[]` when a process exited 0 with no stdout and no stderr. `complete_module` then saw `failed == 0` and proceeded to export. A student whose notebook is empty, has all code commented out, or whose `if __name__ == '__main__'` block is missing would produce exactly this: Python exits 0 with no output.

Fixed by adding a failure entry when returncode is 0 but there is no output at all, with a message pointing the student toward the missing test output.

## Test plan

- [ ] `tito module complete 01` now runs `tests/01_tensor/test_01_tensor_progressive.py` in step 3
- [ ] A module with an empty notebook exits 1 at step 1 instead of silently passing
- [ ] Modules with no progressive test file still skip gracefully with a dim warning
- [ ] Normal passing notebooks are unchanged

### DIFF
--- a/tinytorch/tito/commands/module/workflow.py
+++ b/tinytorch/tito/commands/module/workflow.py
@@ -942,17 +942,35 @@ class ModuleWorkflowCommand(BaseCommand):
             'returncode': result.returncode
         }
 
+    def _find_integration_test_file(self, module_name: str) -> Optional[Path]:
+        """Find the progressive integration test file for a module.
+
+        The runner used to look only for test_progressive_integration.py, which
+        does not exist in any module. The actual files follow the pattern
+        test_NN_<name>_progressive.py (e.g. test_01_tensor_progressive.py).
+        This method finds that file so integration tests actually run.
+        """
+        test_dir = Path.cwd() / "tests" / module_name
+        if not test_dir.exists():
+            return None
+        module_num = module_name.split("_")[0]
+        for candidate in sorted(test_dir.glob("*.py")):
+            name = candidate.name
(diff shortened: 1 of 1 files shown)


profvjreddi:
Both bugs are real and both are fixed on dev (`16b246fd84`), @Shashank-Tripathi-07. The integration tests really were looking for `test_progressive_integration.py`, which exists in no module, so they never ran during `module complete`; that now resolves `test_<module>_progressive.py` with a glob fallback. And the notebook syntax check from #1877 runs before export.

On how `module complete` should behave, I kept it additive: the `src/` unit tests still run, and the syntax check is an extra gate rather than a replacement. Closing as landed, with your authorship on the commit. Two good finds.

PR #1894, opened 2026-06-24
fix(book): render pseudocode algorithms and contain their overflow

farhan523:
## Summary

Fixes two HTML-output bugs affecting **every algorithm block across both volumes** (first surfaced on `vol1/nn_computation`).

### 1. Algorithms render as raw LaTeX instead of typeset pseudocode
pseudocode.js's init script ran synchronously at parse time, but the book renders math with **MathJax loaded via `defer`** and no KaTeX. So at the moment `renderElement` ran, no math backend existed yet and it threw:

```
Uncaught EvalError: No math backend found. Please setup KaTeX or MathJax.
```

The raw `\begin{algorithm}…` source was left on the page and never retried (MathJax loads moments later and typesets the rest of the page, but the algorithm init had already failed). **Reproduced on the live `_dev` site**, so this is a production bug, not a local artifact.

**Fix:** wait for the math backend to be ready before rendering — KaTeX if present synchronously, otherwise MathJax's `startup.promise` (available after the deferred MathJax script runs).

### 2. Wide algorithms force page-level horizontal scroll
The rendered `.pseudocode-container` had no overflow handling, so a long algorithm line (e.g. the `\Require …` signature) pushed the whole page sideways. Scoped `overflow-x: auto` to the container so it scrolls in its own box — same failure mode and fix as the existing wide-table rule directly above it.

## Verification
Rendered the page and drove headless Chrome against the output:

| | Before |
Screenshot 2026-06-24 105451

| After |
Screenshot 2026-06-24 105441

MathJax confirmed working independently (847 `mjx-container` elements rendered), isolating the failure to pseudocode init timing.

## Scope
Both files are shared across all chapters and both volumes, so this repairs every algorithm block — not just the one page.

### DIFF
--- a/book/quarto/_extensions/mlsysbook-ext/pseudocode/pseudocode.lua
+++ b/book/quarto/_extensions/mlsysbook-ext/pseudocode/pseudocode.lua
@@ -24,34 +24,53 @@ local function ensure_html_deps()
     [[
     <script type="text/javascript">
     (function(d) {
-      d.querySelectorAll(".pseudocode-container").forEach(function(el) {
-        let pseudocodeOptions = {
-          indentSize: el.dataset.indentSize,
-          commentDelimiter: " " + el.dataset.commentDelimiter + " ",
-          lineNumber: el.dataset.lineNumber.toLowerCase() === "true",
-          lineNumberPunc: el.dataset.lineNumberPunc,
-          noEnd: el.dataset.noEnd.toLowerCase() === "true",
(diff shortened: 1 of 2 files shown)


profvjreddi:
You rock! Thanks for catching this @farhan523 👍 This was a new addition, and I forgot to check the rendering.

farhan523:
> You rock! Thanks for catching this @farhan523 👍 This was a new addition, and I forgot to check the rendering. 

I was exploring the book, saw it and fixed this bug. 🙌 Glad we caught it before release.

PR #1902, opened 2026-06-26
fix(staffml): guard localStorage access in ThemeProvider against strict-storage browsers

farhan523:
## Problem

`ThemeProvider.getInitialTheme()` read `localStorage` without a `try/catch`, while the matching writer (`toggleTheme`) was already guarded. In browsers that block storage — Safari with **"Block all cookies"**, or sandboxed iframes without `allow-same-origin` — even accessing `window.localStorage` throws a `SecurityError` (not just the `.getItem()` call). That throw happened inside the mount `useEffect`, aborting it **before** `setTheme` ran, so the theme never reconciled and stayed stuck at the SSR default.

## Fix

Funnel all storage access through shared `safeGet` / `safeSet` helpers, so reads and writes share a single guard and a future caller can't reintroduce an unguarded access that breaks a mount effect. Runtime validation of the stored value (`=== "light" | "dark"`) is preserved, so a garbage value can't poison the theme.

## Test

Adds `theme-provider-storage.test.tsx`:
- mounts `ThemeProvider` with a throwing `getItem` → no throw, falls back to `light`
- confirms normal reconciliation to a stored `dark` theme still works

No behavior change for users with working storage.

### DIFF
--- a/interviews/staffml/src/__tests__/theme-provider-storage.test.tsx
+++ b/interviews/staffml/src/__tests__/theme-provider-storage.test.tsx
@@ -1,0 +1,54 @@
+/**
+ * ThemeProvider storage-guard regression.
+ *
+ * getInitialTheme() reads localStorage to reconcile the stored theme on mount.
+ * In strict-storage browsers (Safari "Block all cookies", sandboxed iframes)
+ * localStorage.getItem itself throws. The writer (toggleTheme) already guarded
+ * its setItem calls, but the reader did not — so the mount effect threw and the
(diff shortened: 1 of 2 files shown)


profvjreddi:
@all-contributors please add @farhan523 for code, bug

profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). The localStorage guard correctly covers the property access throw, not just the getItem call, so the mount effect reconciles the theme even in strict storage browsers. Credited via all-contributors.

PR #1903, opened 2026-06-26
fix(staffml): expose expanded state on Nav disclosure toggles

farhan523:
## Problem

The **Tools** dropdown toggle and the **mobile hamburger** in the top nav are disclosure buttons that show/hide a menu, but they exposed no expanded/collapsed state to assistive technology. The only cue that the menu was open was a visual chevron rotation, so screen-reader and keyboard users had no programmatic signal (WCAG 4.1.2).

## Fix

- Add `aria-haspopup="true"`, `aria-expanded={open}`, and `aria-controls` to both toggles.
- Give each controlled menu a matching `id` (`nav-tools-menu`, `nav-mobile-menu`).
- Mark the purely decorative chevron / menu / close icons `aria-hidden="true"`.

No visual or behavioral change for sighted users.

## Test

Adds `nav-disclosure-aria.test.tsx` asserting both toggles report `aria-expanded` correctly before/after click and that `aria-controls` points at the rendered menu.

### DIFF
--- a/interviews/staffml/src/__tests__/nav-disclosure-aria.test.tsx
+++ b/interviews/staffml/src/__tests__/nav-disclosure-aria.test.tsx
@@ -1,0 +1,60 @@
+/**
+ * Nav disclosure ARIA guardrails.
+ *
+ * The "Tools" dropdown toggle and the mobile hamburger are disclosure buttons
+ * that show/hide a menu, but they exposed no expanded/collapsed state to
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). aria-expanded plus aria-controls on both nav disclosures, with the decorative chevron and menu icons hidden from assistive tech. Credited via all-contributors.

PR #1904, opened 2026-06-26
fix(staffml): give StarGate proper dialog semantics and focus management

farhan523:
## Problem

`StarGate` is a full-viewport **blocking** overlay (the "only ask" star gate), but it rendered as a plain `<div>`:

- no `role="dialog"`, `aria-modal`, or `aria-labelledby` — screen readers didn't announce it as a dialog
- no Escape handler
- no focus management — focus stayed on whatever was behind the gate, and Tab walked into the dimmed, non-interactive page behind the overlay (WCAG 2.4.3 / 2.1.2 / 4.1.2)

## Fix

Mirror the pattern already used by `KeyboardShortcutsOverlay` and `CommandPalette`:

- `role="dialog"` + `aria-modal="true"` + `aria-labelledby` pointing at the heading (both the gate and the post-action "Thank you" confirmation panel)
- move focus to the primary "Star on GitHub" CTA on mount, restore focus to the previously-focused element on unmount
- trap Tab/Shift+Tab within the dialog surface
- Escape dismisses the gate (honor-system: counts as a dismiss, same as the ✕)

No visual change.

## Test

Adds `stargate-dialog-a11y.test.tsx` covering dialog attributes + labelling, focus-to-CTA on mount, Escape→dismiss, and focus restoration on unmount. Full suite (117 tests) green.

### DIFF
--- a/interviews/staffml/src/__tests__/stargate-dialog-a11y.test.tsx
+++ b/interviews/staffml/src/__tests__/stargate-dialog-a11y.test.tsx
@@ -1,0 +1,63 @@
+/**
+ * StarGate dialog accessibility guardrails.
+ *
+ * StarGate is a full-viewport blocking overlay, but it used to render as a
+ * plain <div>: no role="dialog", no aria-modal, no aria-labelledby, no Escape
+ * handler, and no focus management — so the dimmed page behind it stayed
+ * keyboard-reachable and screen readers didn't announce it as a dialog.
+ *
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). Proper dialog semantics, focus trap, focus restore, and Escape handling for the StarGate overlay, matching the existing overlay pattern. Credited via all-contributors.

PR #1905, opened 2026-06-26
fix(staffml): dialog semantics + focus management for framework primitive-detail modal

farhan523:
## Problem

The framework page's `PrimitiveDetail` overlay rendered as a plain `<div>`:

- no `role="dialog"`, `aria-modal`, or `aria-labelledby` — screen readers didn't announce it as a dialog (WCAG 4.1.2)
- no focus management — focus stayed in the dimmed page behind the modal, and Tab walked through the hidden content (WCAG 2.4.3)

(Escape-to-close was already handled by the page-level keydown listener.)

## Fix

Mirror the pattern already used by `KeyboardShortcutsOverlay` / `CommandPalette`:

- `role="dialog"` + `aria-modal="true"` + `aria-labelledby` pointing at the primitive name
- move focus to the close button on open, restore focus to the previously-focused element on close
- trap Tab/Shift+Tab within the panel

`PrimitiveDetail` is exported so it can be unit-tested in isolation. No visual change.

## Test

Adds `framework-modal-a11y.test.tsx` covering dialog attributes + labelling, focus-to-close on mount, and focus restoration on unmount.

### DIFF
--- a/interviews/staffml/src/__tests__/framework-modal-a11y.test.tsx
+++ b/interviews/staffml/src/__tests__/framework-modal-a11y.test.tsx
@@ -1,0 +1,44 @@
+/**
+ * Framework primitive-detail modal accessibility guardrails.
+ *
+ * The PrimitiveDetail overlay rendered as a plain <div>: no role="dialog",
+ * aria-modal, or aria-labelledby, and no focus management — so screen readers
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). Dialog semantics and focus management for the framework primitive modal. Credited via all-contributors.

PR #1906, opened 2026-06-26
fix(staffml): remove invalid HTML button nested inside the explore SVG

farhan523:
## Problem

The radial explorer's center "zoom out" affordance was an HTML `<button>` wrapping an SVG `<circle>`:

```jsx
<button type="button" onClick={...} className="cursor-pointer">
  <circle cx={CX} cy={CY} r="92" ... />
</button>
```

An HTML `<button>` is **not valid SVG content** — it doesn't focus reliably across browsers and exposed no accessible name (WCAG 4.1.2). At the root focus it was also a silent no-op.

## Fix

- Move the click handler onto the `<circle>` directly (valid SVG, preserves the mouse zoom-out).
- Only show the pointer/hover affordance when there is a parent focus to zoom out to.

Keyboard navigation is unchanged: the **Breadcrumb** above the chart and the **ExplorerPanel** beside it already provide real `<button>` controls for the same focus changes, and the chart is exposed to assistive tech as a single labelled `role="img"`. A code comment documents this so the redundancy is intentional and discoverable.

## Test

Adds `explore-svg-markup.test.tsx` asserting the chart renders as a labelled `role="img"` with no nested HTML `<button>`.

### DIFF
--- a/interviews/staffml/src/__tests__/explore-svg-markup.test.tsx
+++ b/interviews/staffml/src/__tests__/explore-svg-markup.test.tsx
@@ -1,0 +1,26 @@
+/**
+ * Explore radial-chart markup guardrail.
+ *
+ * The center "zoom out" affordance used to be an HTML <button> wrapping an SVG
+ * <circle>. A <button> is not valid SVG content — it doesn't focus reliably
+ * across browsers and has no accessible name. The click handler now lives on
+ * the <circle> directly; the keyboard-accessible navigation is the Breadcrumb
+ * and the ExplorerPanel, and the chart is exposed to assistive tech as a
+ * single labelled role="img".
+ *
+ * This test locks in that the SVG contains no nested HTML <button>.
+ */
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). Removing the invalid HTML button nested in the SVG is the right call, and I confirmed the Breadcrumb keeps the keyboard path for zoom out, so accessibility improves on net. Credited via all-contributors.

profvjreddi:
Hi @farhan523, good question. Your changes are all in dev. I integrated that batch locally (cherry-picked your commits onto a review branch, verified CI, then merged, see 54536f6d and its siblings from June 26, all with your authorship). GitHub only shows the purple "merged" badge when the PR's exact head commit reaches the base branch, and a cherry-pick creates a new SHA, so these show "closed" instead. Nothing was lost, your credit and authorship are intact. Your more recent PRs were merged directly on GitHub, which is why those show "merged".

PR #1907, opened 2026-06-26
fix(staffml): track and clear toast auto-dismiss timers

farhan523:
## Problem

Each toast schedules a 4s `setTimeout` to auto-dismiss itself, but the timer handle was discarded:

```js
const show = useCallback((msg) => {
  const id = nextId++;
  setToasts(prev => [...prev, { ...msg, id }]);
  setTimeout(() => { setToasts(prev => prev.filter(t => t.id !== id)); }, 4000); // never cleared
}, []);
```

So the timer was never cleared. If `ToastProvider` unmounts within 4s of a toast (route teardown, fast-refresh, test cleanup) the timer still fires `setToasts` on a dead tree, and a manual dismiss leaves the timer running to a wasted no-op tick.

## Fix

- Track each toast's timer in a `useRef<Map<id, timeout>>`.
- Cancel the timer in `dismiss()`.
- Clear all pending timers in a provider unmount cleanup.

No user-visible behavior change.

## Test

Adds `toast-timer-cleanup.test.tsx` (fake timers) asserting the pending timer is cleared on unmount and on manual dismiss, and that flushing timers afterward doesn't throw.

### DIFF
--- a/interviews/staffml/src/__tests__/toast-timer-cleanup.test.tsx
+++ b/interviews/staffml/src/__tests__/toast-timer-cleanup.test.tsx
@@ -1,0 +1,77 @@
+/**
+ * Toast auto-dismiss timer cleanup regression.
+ *
+ * Each toast schedules a 4s setTimeout to auto-dismiss itself. The timer
+ * handle used to be discarded, so it was never cleared: if the provider
+ * unmounted within 4s the timer still fired setState on a dead tree, and a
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks @farhan523. Integrated into dev (CI green). Tracking and clearing the toast auto dismiss timers fixes the setState after unmount hazard, on both manual dismiss and provider unmount. Credited via all-contributors.

PR #1913, opened 2026-06-28
fix(staffml): cancel StarGate's verify-delay timer on unmount

farhan523:
## Summary
Both `handleStar` (\"Star on GitHub\") and `handleAlreadyStarred` (\"I already starred\") fired `setTimeout(onVerified, 1500)` to let the user see the \"Thank you\" confirmation panel before the parent dismisses the gate. The returned timer id was never stored, so if the gate unmounted within that 1.5s window — parent route change, programmatic close, or anything that removes `StarGate` from the tree — the timer kept running and `onVerified` fired later on an unmounted parent.

Today the only caller (`practice/page.tsx`) treats a stray `onVerified` call as a no-op (it just sets `showStarGate=false` which was already false), so this was a **latent** leak, not a visible bug. But the contract becomes accidentally correct rather than correct-by-design — a future caller where `onVerified` mutates state on a stale tree would surface it immediately.

### Changes
- Store the timer id in `verifyTimeoutRef` (mirrors the focus-timer pattern already in the same file).
- Add a no-deps cleanup `useEffect` that clears `verifyTimeoutRef.current` on unmount.
- Both handlers assign into the same ref, which also implicitly handles the unlikely back-to-back click case.

## Test plan
- [x] `npx vitest run` → 129/129 across 23 files passing (was 126; +3 new for unmount-after-Star, unmount-after-Already-starred, and still-mounted-fires-at-1.5s).
- [x] `npx tsc --noEmit` → clean.
- [ ] Manual repro (the regression isn't user-visible today; the test is the verification surface).

### DIFF
--- a/interviews/staffml/src/__tests__/stargate-timer-cleanup.test.tsx
+++ b/interviews/staffml/src/__tests__/stargate-timer-cleanup.test.tsx
@@ -1,0 +1,74 @@
+/**
+ * StarGate timer-cleanup regression.
+ *
+ * Both `handleStar` (Star on GitHub) and `handleAlreadyStarred` (I already
+ * starred) schedule `setTimeout(onVerified, 1500)` to give the user a
+ * moment to see the "Thank you" confirmation before the parent dismisses
+ * the gate. If the gate unmounts before 1500ms (parent route change,
+ * programmatic close), the timer should be cancelled — otherwise it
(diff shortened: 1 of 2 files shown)


profvjreddi:
@all-contributors please add @farhan523 for bug, code, test in staffml

PR #1944, opened 2026-07-07
feat(tinytorch): make the website an installable PWA

farhan523:
## Summary

Makes the TinyTorch site (mlsysbook.ai/tinytorch/) an installable Progressive Web App with offline reading support.

- **`manifest.webmanifest`** — installability metadata with relative `start_url`/`scope` so it resolves correctly both under `/tinytorch/` in production and in local previews served at the root. Includes `any` + `maskable` icons (192/512) generated from the flame mark in `logo-tinytorch-simple.png`, themed with the existing `#D4740C` brand color.
- **`sw.js`** (service worker) — conservative caching tuned for a frequently-deployed docs site:
  - **Pages: network-first.** A deploy is picked up immediately; the cache is only an offline fallback, with a branded offline page as the last resort.
  - **Same-origin static assets (css/js/images/fonts): stale-while-revalidate**, capped at 200 entries.
  - **JSON/data passthrough** — `release-manifest.json`, `search.json`, and any fetch/XHR always hit the network, so the release pill and search never go stale.
  - Cross-origin CDN requests (Google Fonts, cdnjs) are left to the browser.
- **`assets/scripts/pwa.html`** — registers the worker at the site base path, derived at runtime because pages render at several directory depths and the site root differs between production (`/tinytorch/`) and previews (`/`). Registration is skipped silently on insecure contexts.
- **apple-touch-icon fix** — the previous href pointed at the 583×328 wordmark; iOS home screens now get a proper 180×180 icon. The manifest `<link>` and icon use leading-`/` project-relative hrefs, which Quarto relativizes per page depth.

No changes to page content or build tooling; `quarto render` output just gains the manifest, worker, and icons via the existing `resources:` mechanism, so the publish workflow deploys them as-is.

## Testing

- `quarto render` passes; verified the manifest link and icon hrefs relativize correctly at all page depths (`index.html`, `modules/`, `tito/`).
- Served `_build` locally: all PWA URLs return 200, manifest parses, `node --check sw.js` passes.
- Verified in Chrome: install prompt appears, worker activates with site-wide scope, visited pages load offline, unvisited pages show the offline fallback.


Shashank-Tripathi-07 (commented):
Hey @farhan523 . 

Totally like and appreciate the idea of bringing TinyTorch from the current infra to a PWA, but we're currently focusing on stabilizing and making the current TinyTorch system so as to prepare for it's major release and hence, I can't merge this PR and I've to close it. You can instead focus on surgical changes and changes that make TinyTorch better for release rather than pushing major changes to TinyTorch for now. Thanks !

PR #1945, opened 2026-07-07
feat(staffml): make the app an installable PWA

farhan523:
## Summary

Makes StaffML installable as a Progressive Web App with offline support — a natural fit for an app that already runs 100% client-side.

- **`src/app/manifest.ts`** — Next.js manifest route (`force-static` so it works with `output: export`). `start_url`/`scope`/`id`/icon URLs carry `NEXT_PUBLIC_BASE_PATH` explicitly, following the same convention as the og-image URLs in `layout.tsx`. Theme/background match the light default from `theme-bootstrap.js`.
- **`public/icons/`** — `any` + `maskable` icons (192/512) plus a 180×180 `apple-touch-icon`, rasterized from the roofline mark in `favicon.svg`. Maskable/apple variants are full-bleed so Android's circle/squircle masks and iOS's corner radius never clip the mark. `layout.tsx` metadata now links the apple icon (iOS ignores SVG favicons).
- **`public/sw.js`** — new, clearly-delimited **app-shell section** appended below the existing vault-API cache. The two sections never compete for a request: the vault handler exits early unless the URL starts with the vault API origin (cross-origin); the app-shell handler exits early unless the request is same-origin *and inside the worker's scope* (so the surrounding Quarto book's assets are left alone). Strategy:
  - **Navigations: network-first** — a deploy is picked up immediately; cache only serves offline revisits, with an inline branded offline page as last resort.
  - **script/style/image/font: stale-while-revalidate**, 300-entry cap (`_next/static/*` is content-hashed so staleness is impossible there).
  - **Everything else passes through** — corpus JSON, XHR/fetch data freshness stays owned by the vault section and `VersionDriftToast`.
  - Cache prefix `staffml-app-` is disjoint from `staffml-vault-`, so the vault section's release-keyed pruning and the app section's version pruning can never delete each other's caches. No changes to any existing vault logic.
- **`corpus-provider.tsx`** — the service worker now registers **unconditionally** instead of only after a successful vault probe: offline support must not depend on the vault API being reachable, since offline is exactly when the probe fails. The `SET_VAULT_API_ORIGIN` message stays gated on probe success, so the vault section stays inert until then (preserving the Chip R4-C-2 message-hardening path unchanged).

## Regression test

`src/lib/__tests__/pwa-manifest.test.ts` guards the manifest↔icons contract: the manifest declares the fields installability requires (`name`, `short_name`, `display: standalone`, scope containing start_url, both `any` and `maskable` at 192+512), and every icon it references exists in `public/` with exactly the pixel dimensions it advertises (PNG IHDR check) — a mismatch there doesn't fail the build, it just silently breaks installability in the field.

## Testing

- `vitest`: 133/133 passing (7 new).
- `next build` clean **both with and without** `NEXT_PUBLIC_BASE_PATH=/staffml`; verified exported HTML links the manifest and apple-touch-icon with the correct prefix in both configurations, and the manifest's `start_url`/`scope`/icon URLs all carry the base path.
- Served the export locally and verified in Chrome: install prompt appears, worker activates, visited pages load offline, unvisited pages get the branded offline fallback, and the vault section still receives `SET_VAULT_API_ORIGIN` after a successful probe.
- Note: `npm run lint` currently crashes on `dev` as well (ESLint 10 / `eslint-plugin-react` incompatibility inside `eslint-config-next`, triggered on `analytics-worker/worker.js`) — pre-existing and unrelated; scoped lint can't run either since the crash happens at rule-load time.


Shashank-Tripathi-07:
Hi @farhan523 ! 

Thanks for the PR, I've gone through it and feel that @profvjreddi will be the one to advise us on this change as it is quite heavy and requires us to implement new system design changes. Imo, electron apps take more hardware capacity than the current model we have [this is based on my analysis and knowledge of PWA]. As our focus is to make the project light enough that anyone with basic hardware can run it, the PWA can be a potential blocker and hence, I'd like prof. vijay to give his review about this before we proceed.

farhan523:
> Hi @farhan523 ! 
> 
> Thanks for the PR, I've gone through it and feel that @profvjreddi will be the one to advise us on this change as it is quite heavy and requires us to implement new system design changes. Imo, electron apps take more hardware capacity than the current model we have [this is based on my analysis and knowledge of PWA]. As our focus is to make the project light enough that anyone with basic hardware can run it, the PWA can be a potential blocker and hence, I'd like prof. vijay to give his review about this before we proceed. 

Hi @Shashank-Tripathi-07 , thanks for the feedback, I am not using electron js, to make this PWA. Electron framework access resorces directly from OS and behave as a native desktop APP. PWA is different from native desktop or mobile apps. It uses browser Navigator API. Opening PWA is similar to opening a new browser tab, and PWA run directly on chrome or other browser from which it is installed, so same amount of resources are used, as opening the link on any  browser's tab.

PR #1948, opened 2026-07-07
a11y(shared): make subscribe modal a real dialog for keyboard and screen-reader users

farhan523:
## Problem

The shared subscribe modal (`shared/scripts/subscribe-modal.js`, mirrored to the book, site, labs, kits, and mlsysim subsites) is a plain `<div>` overlay with several accessibility gaps:

- **Not announced as a dialog** — no `role="dialog"`, `aria-modal`, or accessible name. Screen-reader users get no announcement when it opens and can keep reading the page *behind* the overlay without knowing a form is blocking the view.
- **No focus trap** — a few Tab presses inside the form move focus out of the modal into the visually hidden background page; the focus outline disappears behind the overlay (WCAG 2.4.3).
- **No focus restore** — closing the modal drops keyboard users at the top of the document instead of back on the Subscribe link they activated (WCAG 4.1.2 context).
- **Motion regardless of preference** — the slide-up/scale entrance plays even with `prefers-reduced-motion: reduce` set (WCAG 2.3.3).

## Fix (canonical file only, mirrors regenerated)

- `role="dialog" aria-modal="true" aria-labelledby="subscribe-modal-title"` on `.modal-container`, id added to the existing title → announced as "Stay in the Loop, dialog".
- Tab/Shift+Tab cycle within the modal's visible, enabled controls. The focusable list is computed per keypress, so the trap also works on the post-submit success view where the form is hidden.
- `openModal()` saves `document.activeElement`; `closeModal()` refocuses it.
- `@media (prefers-reduced-motion: reduce)` disables the `fadeIn`/`slideUp` animations.

## Behavior

No visual or functional change for mouse users: same design, same animations (when motion is allowed), Escape and overlay-click still close it.

Mirrors were regenerated with `bash shared/scripts/sync-mirrors.sh` (all six copies stay byte-identical for the CI drift check). Verified with `node --check`.

### DIFF
--- a/book/quarto/assets/scripts/subscribe-modal.js
+++ b/book/quarto/assets/scripts/subscribe-modal.js
@@ -28,14 +28,14 @@
   function createModalHTML() {
     return `
       <div id="subscribe-modal" class="modal-overlay" style="display: none;">
-        <div class="modal-container">
+        <div class="modal-container" role="dialog" aria-modal="true" aria-labelledby="subscribe-modal-title">
           <button class="modal-close" data-close-modal aria-label="Close">&times;</button>
           <div class="modal-content">
             <div class="modal-header">
               <div class="modal-brand-row">
                 <span class="modal-brand-item">📚 MLSysBook</span>
               </div>
-              <h2 class="modal-title">Stay in the Loop</h2>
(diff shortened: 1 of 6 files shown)


profvjreddi:
Thank you for the work on the shared subscribe modal, @farhan523. Merged, and much appreciated.

PR #1951, opened 2026-07-07
fix(tinytorch): integration test collection failure silently treated as no-tests

Shashank-Tripathi-07:
## Summary
- `_run_integration_tests()` captures `result.returncode` from the pytest subprocess but no caller (`complete_module`, `_complete_module_quiet`, `run_module_tests`) ever reads it -- only `'passed'`/`'failed'` are checked.
- If pytest fails during **collection** (e.g. an import-time error in the exported package -- distinct from the notebook-syntax check, which only `compile()`s notebook cells, not the exported package), it exits nonzero with no `PASSED`/`FAILED` lines to parse. `_parse_pytest_output` returns an empty list, so the result comes back as `passed=0, failed=0`.
- Every caller's `if integration_result['failed'] > 0` check is then `False`, so `tito module complete` prints `"No integration tests for this module"` and marks the module complete despite the exported package being unimportable/broken.

## Fix
Treat "zero parsed tests + nonzero pytest exit code" as a collection failure rather than "no tests exist", surfacing it through the same tests-list/error-field shape `_parse_test_output` already uses for its own unit-test fallback case. The legitimate "no test file for this module" short-circuit (returns before pytest even runs) and the "pytest ran fine but collected zero tests" case (`returncode == 0`) are unaffected.

## Test plan
- [x] Fed `_parse_pytest_output` a real pytest collection-error transcript (`ImportError` during test collection): confirmed it returns zero parsed tests, matching the bug's premise.
- [x] Confirmed the new guard turns that into `failed=1` instead of the old `passed=0/failed=0`, which now correctly trips the existing `if integration_result['failed'] > 0` branch in every caller.

### DIFF
--- a/tinytorch/tito/commands/module/workflow.py
+++ b/tinytorch/tito/commands/module/workflow.py
@@ -1037,14 +1037,28 @@ class ModuleWorkflowCommand(BaseCommand):
             # pytest itself errored (e.g. a collection-time import failure in
             # the exported package) rather than legitimately having zero
             # tests. Surface this as a failure instead of silently reporting
-            # "no integration tests for this module".
+            # "no integration tests for this module" -- except for two cases
+            # that are not real collection failures:
+            #   - exit code 5: pytest's own "no tests collected" signal
+            #   - exit code 4: a pytest.UsageError, which conftest.py raises
+            #     from _validate_package_exported() when core modules that
+            #     come *later* in the build order haven't been exported yet.
+            #     That check is unconditional (it requires every core file to
+            #     exist, not just the one under test), so it legitimately
+            #     trips for early modules during a progressive build.
             error_msg = (result.stderr or result.stdout).strip()
-            concise_error = '\n'.join(error_msg.split('\n')[:5]) if error_msg else "pytest exited with an error"
-            tests_run = [{
-                'name': 'pytest_collection',
-                'passed': False,
-                'error': concise_error,
-            }]
+            is_no_tests_collected = result.returncode == 5
+            is_progressive_export_gate = (
+                result.returncode == 4
+                and "TINYTORCH PACKAGE NOT EXPORTED" in error_msg
+            )
+            if not is_no_tests_collected and not is_progressive_export_gate:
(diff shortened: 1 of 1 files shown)


profvjreddi:
Good catch, @Shashank-Tripathi-07 — you're right that `_run_integration_tests` captures the pytest returncode but no caller reads it, so a nonzero exit with no parsed PASSED/FAILED lines gets swallowed as 'no tests' and the module is still marked complete. That hole is real and worth closing.

Holding it rather than merging, though, because the fix as written turns the normal progressive Inline Build red (which is what your CI is showing): module 01 runs its integration tests before modules 02–04 have exported their files, so `conftest.py`'s `_validate_package_exported()` raises a `pytest.UsageError` (exit 4, zero parsed tests) and the new synthetic node fails on it. Could you (a) exclude pytest exit 5 ('no tests collected') and the conftest's progressive-export `UsageError` — or relax the conftest so it only requires the module under test, not future ones — and (b) confirm the full `tito` progressive build stays green? Then it's a clean land. Thanks for digging into this, and sorry for the slow review.

Shashank-Tripathi-07:
Thanks for the detailed review, and no worries on the timing.

You're right, the guard was too broad. I've scoped it down to only flag genuine collection failures, excluding two legitimate non-failure signals:
- pytest exit code 5 (`no tests collected`)
- pytest exit code 4 (`UsageError`), specifically when it's conftest's `_validate_package_exported()` progressive-export gate (matched via the `TINYTORCH PACKAGE NOT EXPORTED` marker in its error message), which unconditionally requires every core file to exist, so it's expected to trip on early modules before later ones have exported.

A real collection failure (e.g. a broken import in the module's own exported code) still surfaces correctly, since it won't carry that marker or exit code 5.

Confirmed the full progressive build stays green: ran `tito dev test --inline --ci` locally (20/20 modules passed), and CI's Stage 1 Inline Build now passes on both ubuntu-latest and windows-2022, with everything downstream green too.

Ready for another look whenever you have time.

Issue #1792, opened 2026-05-21
StaffML: Improvement suggestions — MCQ mode, coding questions, recent-papers section, and on-device architecture content


profvjreddi:
## Summary

Feedback from an external user on **StaffML** (https://mlsysbook.ai/staffml/welcome/), currently described as "similar to LeetCode for MLSys" with concept coverage that is "crystal & clear." The suggestions below are additive enhancements to broaden the platform's usefulness for students and engineers preparing for MLSys interviews.

## Suggested improvements

- [ ] **MCQ answer format** — Offer a multiple-choice format alongside the current descriptive answers. Many hiring platforms screen candidates with MCQs, so an MCQ mode would let students and engineers practice in the format they will actually face.
- [ ] **Hands-on coding questions (Python / C++)** — Add coding exercises. Companies frequently ask candidates to implement an optimization in real time during interviews, so live-coding practice would meaningfully increase the platform's value.
- [ ] **Recent-papers section** — Add a small, curated section highlighting recent high-impact papers so users stay current with the field. Example: TurboQuant (Zandieh et al., Google Research; arXiv:2504.19874), which has drawn notable attention.
- [ ] **On-device deployment & emerging-architecture content** — Add Q&A and blog posts on deploying and optimizing newer architectures on-device, e.g. Mixture-of-Experts (MoE) models. Extending beyond core concepts into applied, cutting-edge deployment would broaden the platform's reach.

## Notes

The same reviewer also gave positive feedback on the main MLSys book update — the expanded device-deployment and systems-side material was called a valuable, well-documented reference that colleagues working in MLSys had been lacking.

*Filed on behalf of feedback received from an external user.*

DIYA73:
I'd like to pick up the recent-papers section specifically — starting with a small curated list, structured so it's easy to swap in automated Alphaxiv/arXiv pulling later per @Shashank-Tripathi-07's suggestion. Will open a PR against dev soon.

Shashank-Tripathi-07:
Hello @DIYA73 , I'd like you to ask for permission from prof. before building a PR. Going straight for a PR without discussing a feature can lead to the PR rejection and that can lead to waste of efforts.

DIYA73:
@profvjreddi — following up in case this got buried. Would a small curated recent-papers section (manual list to start, per the plan above) be a reasonable place to start on this? Happy to open the PR once you give the go-ahead.

profvjreddi:
[...]

PR #1962, opened 2026-07-22
feat(staffml): add recent-papers section to About page

DIYA73:
one line — Addresses #1792 (recent-papers section only — curated list to start).


interviews/staffml/src/components/RecentPapers.tsx
Shashank-Tripathi-07, line 1: Pretty neat part of the code, how about having a rolling system that would pull in the best papers that have been released recently. That would bring independence in the system and we won't need to replace the papers in here.

interviews/staffml/src/components/RecentPapers.tsx
```diff
@@ -0,0 +1,39 @@
...
+  url: string;
+}
+
+const papers: Paper[] = [
+  {
+    title: "TurboQuant: Online Vector Quantization with Near-optimal Distortion Rate",
+    authors: "Zandieh et al., 2025",
+    url: "https://arxiv.org/abs/2504.19874",
+  },
+  // add 3-5 more here
```
Shashank-Tripathi-07: Please check how the UI updates with multiple papers too. It once happened with me that after adding multiple examples the UI broke and let's avoid it from happening here. Just test it locally and confirm it or you can add multiple papers here too !

Shashank-Tripathi-07 (commented):
Pretty good work, can you go through the comments I made and do the steps and give it an update. Then this will be perfect to merge into the system.

interviews/staffml/src/components/RecentPapers.tsx
DIYA73: Agreed that's the better long-term direction — it's actually what we scoped on #1792 (manual list first, automated Alphaxiv/arXiv pulling later). Keeping this PR small on purpose; the flat {title, authors, url} shape here is deliberately simple so a fetch-based version can swap in later without restructuring the component.

interviews/staffml/src/components/RecentPapers.tsx
DIYA73: Good catch — tested locally with 4 entries now (added 3 more, see the diff) and the list renders and wraps cleanly, no breakage.

Shashank-Tripathi-07 (approved):

interviews/staffml/src/components/RecentPapers.tsx
Shashank-Tripathi-07: Agreed, let's roll out a small PR for now and let's bring a better automated system for later on.

PR #1969, opened 2026-07-28
fix(tinytorch): installer can hang forever with no error on Windows

### LINKED ISSUE #1960: TinyTorch Installation doesn't work
### Area

TinyTorch

### Location

TinyTorch Installation

### Description

I tried running the installaltion script, but I received the following message and TinyTorch was not installed. I have Python 3.13.9, Anaconda distribution.

curl -sSL mlsysbook.ai/tinytorch/install.sh | bash
Activate and verify

cd tinytorch
source .venv/Scripts/activate
tito setup

Checking prerequisites...
✓ Git 2.55.0.windows.2

bash: cd: tinytorch: No such file or directory
bash: .venv/Scripts/activate: No such file or directory
bash: tito: command not found

Thank you.

### Expected Behavior

_No response_

_No response_

Shashank-Tripathi-07:
Closes #1960

## The bug

@gusscruz reported the installer printing `Git OK` and then... nothing. No Python check, no error, no prompt, no exit. My earlier reply on that issue guessed it was `curl | bash` piping into a broken tty. Their follow-up (downloading the script first and running `bash install.sh` directly) hit the exact same freeze, which disproves that theory.

## Root cause

`install.sh` had no timeout anywhere. Its Python-detection code ran `python -c "..."` and just waited. On Windows, `python` commonly resolves to the built-in App Execution Alias placeholder (`%LOCALAPPDATA%\Microsoft\WindowsApps\python.exe`), which Windows registers by default even when no real Python is installed (confirmed present on a machine that also has a separate real Python install). Invoking that placeholder with arguments can silently try to open the Microsoft Store and never return control to the caller. One hung subprocess call there froze the entire installer with zero output -- indistinguishable from the script having crashed.

## The fix

Rather than patch just that one call, I went through every external command in the script that can legitimately hang and gave it a bounded timeout plus a plain-language error explaining the likely cause and what to do:

- Python detection/version checks (the direct cause of #1960)
- `git ls-remote` (dead network / silent firewall drop)
- `git clone` (proxy/VPN blocking the connection instead of rejecting it)
- venv creation (corrupted Python install)
- `pip install` steps (generous 10-minute ceiling -- this guards against a true hang, like a private package index silently waiting on credentials nobody can type in a piped `curl | bash` session, without punishing a legitimately slow-but-working install on a bad connection)
- interactive prompts reading from `/dev/tty` (a tty that exists as a path but never delivers input is indistinguishable from a hang otherwise)

Also folded in Python detection for Windows installs where Python is only exposed as a `python.cmd`/`python.bat` shim (e.g. `uv python install`, pyenv-win) rather than a `.exe` -- Git Bash's bare `command -v python` can't see those, which previously reported "Python not found" even when Python was installed and working.

## Testing

- Simulated a hung Python launcher (a fake `python` that never returns): old script hangs forever; patched script fails clearly in ~7s with an actionable message pointing at Windows' App execution aliases setting.
- No Python available at all: clear, fast, actionable error (~7s, one timeout window).
- Simulated an unroutable network call: timeout wrapper reliably kills it and returns within the configured window.
- Full real install end-to-end (git clone -> venv -> pip install -> verify): succeeds, unaffected by the added guards.
- `bash -n` syntax check passes.

### DIFF
--- a/tinytorch/quarto/install.sh
+++ b/tinytorch/quarto/install.sh
@@ -80,6 +80,22 @@ NON_INTERACTIVE="${TINYTORCH_NON_INTERACTIVE:-}"
 TINYTORCH_VERSION="${TINYTORCH_VERSION:-}"
 
 # ============================================================================
+# Timeouts
+# ============================================================================
+# Every external command below (Python detection, network calls, downloads,
+# package installs, interactive prompts) is capped by one of these. None of
+# them should ever be reached in a healthy environment; they exist purely so
+# that a broken environment produces a clear, timed-out error message instead
+# of an installer that just sits there forever with no output and no way to
+# tell what's wrong. See https://github.com/harvard-edge/cs249r_book/issues/1960
+PYTHON_CHECK_TIMEOUT=5      # Detecting/validating a python command
+NETWORK_CHECK_TIMEOUT=15    # Checking GitHub is reachable
+CLONE_TIMEOUT=120           # Downloading TinyTorch (shallow sparse clone)
+VENV_CREATE_TIMEOUT=60      # Creating the virtual environment
+PIP_INSTALL_TIMEOUT=600     # Installing Python packages (10 min ceiling)
+TTY_READ_TIMEOUT=30         # Waiting on an interactive prompt
+
+# ============================================================================
 # ANSI Color Codes (for terminal output)
 # ============================================================================
 RED='\033[0;31m'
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
Pushed a follow-up commit: instead of just timing out and telling the user to fix PATH/Windows settings themselves when Python isn't found, the installer now actively searches the common places Windows Python installers put python.exe (python.org's default dir, Program Files/Program Files (x86), Anaconda/Miniconda, pyenv-win, uv) plus the Python Launcher (py.exe, which has its own install registry separate from PATH), and uses the first working one it finds.

Every candidate still goes through the existing timeout-guarded version check, so this can't hang and can't pick a broken or too-old Python. If it does find something off-PATH, it says so explicitly (which Python, and where) rather than silently using it.

This covers the very common case of someone who has Python installed (often via Anaconda) but never added it to PATH -- previously indistinguishable from "Python not found."

Shashank-Tripathi-07:
Squashed the two commits on this branch into one clean commit (the timeout-guard fix and the off-PATH Python search follow-up were really one change) to keep the history minimal. No functional difference, same diff either way -- just a force-push to clean up the log.

PR #1978, opened 2026-07-31
fix(socratiq): enable AI companion and free Ctrl+/ from search

farhan523:
## Problem

On the [SocratiQ page](https://mlsysbook.ai/contents/frontmatter/socratiq/socratiq.html) the docs tell readers to press **`Ctrl + /`** to open the AI chat. In practice:

1. The SocratiQ bundle is **disabled** in the HTML configs, so nothing loads.
2. Even with it enabled, **`Ctrl + /` opens Quarto's search box**, not the chat.
3. When it does load, the chat sidebar **flashes open-then-closed** on every page load.

## Root causes (traced in code)

- **Disabled bundle:** the `<script … bundle.js>` tag is commented out in `config/_quarto-html-vol1.yml` / `-vol2.yml` ("disabled pending quiz regeneration").
- **Shortcut collision:** Quarto's search (`quarto-search.js`) binds the `/` key on **`keyup`** and checks only `event.key === "/"`, ignoring modifiers — so `Ctrl + /` matches and opens search. SocratiQ's own handler runs on **`keydown`** and calls `preventDefault()`, which cannot stop Quarto's separate `keyup` listener.
- **Load flash:** the slide-out panel ships with `transition-transform` in its **initial** markup. On first paint (before the shadow-root styles apply), the panel renders on-screen, then animates off as styles load — a visible flash.

## Changes

- **Enable SocratiQ** in the vol1/vol2 HTML configs and add `tools/scripts/socratiQ/*.js` as a project resource so the bundle is served.
- **`socratiq-search-shortcut-fix.js`** (new): a capture-phase `keyup` listener that `stopImmediatePropagation()`s `/` **only when Ctrl/Cmd is held**, freeing `Ctrl + /` for SocratiQ while leaving plain `/` search untouched.
- **Flash fix:** remove `transition-transform` from the panel's initial template in `socratiq/src_shadow/index.html` + `indexHtml.js` (and the built `bundle.js` to match). The open/close handlers re-add the class, so animations still work — only the first-paint flash is gone.

## Before → After (for readers)

| | Before | After |
|---|---|---|
| SocratiQ chat | Disabled | Loads & works |
| `Ctrl + /` | Opens search | Opens chat |
| Plain `/` | Search | Search (unchanged) |
| Page load | Sidebar flashes | No flash |

## ⚠️ Maintainer note — please read before merging

This PR **re-enables** a bundle that was intentionally turned off *"pending quiz regeneration."* If the quiz pipeline isn't ready, you may want to **merge only the shortcut + flash fixes and keep the enable out** — happy to split the PR that way. The shortcut and flash fixes are correct and low-risk regardless of when the bundle is enabled.

The `bundle.js` change is a build artifact; the same fix is applied to its source in `socratiq/src_shadow/`, so a rebuild reproduces it.

## Testing

Verified locally with `quarto preview`: `Ctrl + /` opens the chat panel (no search box), plain `/` still opens search, no load flash, and chat streams responses end-to-end via the Cloudflare workers (tested through a local CORS proxy since the workers allowlist `mlsysbook.ai`).


Shashank-Tripathi-07 (commented):
@profvjreddi code looks good to me. Thanks farhan for the contribution. 

 One thing before we merge though, the SocratiQ enable, want your call on it specifically, since you're the one who turned it off (d0a2bd109a, 'pending quiz regeneration'). I Looked into it: the quiz regeneration that blocked it looks done (quiz/vol1-pass/quiz/vol2-pass merged, the last correction  was on 2026-07-16), and SocratiQ's own chat doesn't actually touch the stale *_quizzes.json files anyway, that's inject_quizzes.lua's data, a separate system this PR doesn't re-enable. So it might already be safe, but it was your call to disable it, want your sign-off before we flip it back on. Farhan already offered to split the PR (shortcut + flash fixes only, leave the enable out) if you'd rather do that instead. 

Once you're good with it, I'll tag Kai for a code-level pass too.

dekasthiti:
Thanks, @farhan523. When can we expect the fix to be in main?

farhan523:
Heads up that the `Ctrl + /` fix from this PR **isn't live on staging yet**, and won't be until the bundle enable lands.

The shortcut fix and flash fix merged fine, but the enable was pulled out in 06895ddd5 pending @profvjreddi's sign-off. Without it, `bundle.js` stays commented out in `_quarto-html-vol{1,2}.yml`, so the widget never initializes, `shortcutKeys()` never binds, and `Ctrl + /` still falls through to Quarto's search box. The fix is on `dev` — it just isn't reachable yet.

I've opened #2044 with that one remaining change (4 lines, exactly the inverse of what 06895ddd5 reverted). I've marked it as **not to merge until @profvjreddi signs off**, since it was his call to disable it — it's staged and ready rather than presuming the decision.

@dekasthiti — that's the honest answer to your timing question: it's blocked on that sign-off, not on remaining work.

Shashank-Tripathi-07:
I think this tool was built as a helper for students reading the book. Because socratiq requires an AI API in the backend we can't just allow it and it requires a good release in order to be stabilized. 

However, in your local version of the book and the dev branch. You can run socratiq with your own API Key, this helping you on this. 

I think it'll be merged to main before the next book release but I don't see it coming to active use in some time. All the best @dekasthiti and @farhan523

farhan523:
> I think this tool was built as a helper for students reading the book. Because socratiq requires an AI API in the backend we can't just allow it and it requires a good release in order to be stabilized. 
> 
> However, in your local version of the book and the dev branch. You can run socratiq with your own API Key, this helping you on this. 
> 
> I think it'll be merged to main before the next book release but I don't see it coming to active use in some time. All the best @dekasthiti and @farhan523 

Hi @Shashank-Tripathi-07, backend is already integrated with the chat, when I tried to run the chat locally for testing, I was getting cors error, did some hack to bypass the cors error and chat worked perfectly, I didn't use any key of my own.

Shashank-Tripathi-07:
Pretty cool, need to check this behaviour myself too. Thanks for the report !

PR #2007, opened 2026-08-11
fix(mlsysim): surface load errors via last_load_error on DesignLedger

### LINKED ISSUE #1994: [Bug][Labs] DesignLedger.load()/load_async() silently swallow read failures and reset progress to blank
## Summary

`DesignLedger`'s read path (`load()` for native, `load_async()` for WASM/Pyodide) silently swallows failures and resets to a fresh, empty `LedgerState()`, with no error surfaced anywhere. This is the same class of bug that #1985 fixed on the *write* side (`save_async()`), just on the *read* side, and it currently affects all 34 lab notebooks.

## Relationship to #1985 / #1988

#1985 was "student progress silently lost because `save()` didn't observe or surface IndexedDB failures." #1988 fixed exactly that, on the save path. While reviewing #1988, I found the load path has the identical shape of bug and was deliberately left out of scope for that PR, `save()` was the one actively corrupting/losing new progress, `load()` degrading to a blank state on a transient read failure is a real but distinct failure mode, and bundling both into one PR would have made #1988 harder to review. This issue exists so that decision (and the gap itself) is tracked explicitly rather than silently forgotten once #1988 merges.

## Current behavior (verified against `mlsysim/mlsysim/labs/state.py` on #1988's branch)

**Native path, `load()`** (`state.py:83-104`):
```python
if self.file_path.exists():
    try:
        with open(self.file_path, "r") as f:
            data = json.load(f)
        data["history"] = self._parse_history(data)
        self._state = LedgerState(**data)
    except Exception:
        self._state = LedgerState()
return self._state
```
Any exception, a corrupt JSON file, a permissions error, a partial write, is swallowed with a bare `except Exception`, and the ledger silently resets to a fresh, empty state. Nothing is logged, nothing is raised, nothing distinguishes "no save file exists yet" (a legitimate first-run case) from "a save file exists but couldn't be read" (a real data-loss event).

**WASM path, `load_async()`** (`state.py:106-209`):
```python
except Exception as e:
    print(f"Failed to load from IndexedDB: {e}")
    self._state = LedgerState()
return self._state
```
Same shape: any IndexedDB read failure (a transaction error, `indexedDB.open()` failing, a malformed stored record) is caught, printed to console (not raised, not tracked), and the ledger silently resets to empty. There is no `last_load_error` property mirroring the `last_save_error` that #1988 added for the save path, so calling code has no way to detect that a load failed versus that the student is simply new.

## Blast radius

Every one of the 34 lab notebooks calls this and discards the return value:
```python
_ = await ledger.load_async()
```
(confirmed via `grep -rn "load_async(" labs/vol1 labs/vol2`, 34 matches, one per lab, all with the same `_ = await ...` pattern). Since the return value is thrown away, the *only* way a lab's UI reflects load state is via `self._state` as a side effect. If a read fails for any reason, every lab silently shows the student a blank progress state, with no indication anywhere that this is a failure rather than a fresh start. A student could reasonably conclude they'd never done the lab before, or that a completed lab was reset, with no error, no console signal by default, nothing to report back for support.

## Suggested fix (mirroring #1988's save-side pattern)

- Add a `last_load_error` property (parallel to `last_save_error`) so calling code, and eventually the UI, can detect and surface a failed load instead of silently treating it as "no prior progress."
- Distinguish "no saved data exists" (native: file doesn't exist; WASM: `IndexedDB` query legitimately returns nothing) from "saved data exists but could not be read" (corrupt file, deserialization error, transaction failure). Only the latter should be treated as an error condition; the former is expected behavior for a new student.
- A real-browser regression test mirroring `labs/tests/test_wasm_persistence.py` (added in #1988): seed real IndexedDB with a known-good record, corrupt it, and confirm `load_async()` surfaces the failure via `last_load_error` rather than silently returning an empty ledger.

## Non-goals for this issue

Not proposing to change what happens to the student experience when there's genuinely no prior progress (a fresh `LedgerState()` is correct there), only when a *real* record exists and fails to load.

ShivtejG236:
## Summary

Fixes #1994. `DesignLedger.load()`/`load_async()` previously reset silently
to a blank `LedgerState()` on any read failure (missing file, corrupt JSON,
IndexedDB error), with no way for calling code to distinguish "first run,
nothing saved yet" from "a save exists but couldn't be read" — a silent
data-loss bug on the read path. Adds `last_load_error`, mirroring the
`last_save_error` pattern #1988 established on the write path.

## Area

- [ ] Kits (hardware labs)

(No box above quite fits `mlsysim`/Co-Labs specifically — checking
Infrastructure as the closest match. Actual area: `mlsysim/mlsysim/labs/`
and `labs/tests/`.)

## Changes

- `mlsysim/mlsysim/labs/state.py`: add `_last_load_error` + read-only
  `last_load_error` property; set it in `load()` and `load_async()`'s
  except blocks instead of discarding the exception.
- `mlsysim/tests/test_state.py`: add native-path regression tests
  (missing file is not an error, corrupt file sets `last_load_error` and
  falls back to blank state, error clears on next successful load).
- `labs/tests/test_wasm_persistence.py`: add two real-browser
  (Playwright + Chromium + Pyodide) regression tests for `load_async()` —
  corrupt IndexedDB record, and `indexedDB.open()` throwing synchronously —
  reusing the `served_dir` fixture #1988 introduced.

Rebased cleanly onto `dev` after #1988 merged; no leftover conflicts.

## Testing

- [x] Ran tests (`pytest tests/`)
- [ ] Manual verification (describe below)

Ran explicitly:
```
python -m pytest mlsysim/tests/test_state.py -v      # 11 passed
python -m pytest labs/tests/test_wasm_persistence.py -v   # 3 passed (real Chromium)
```

### DIFF
--- a/labs/tests/test_wasm_persistence.py
+++ b/labs/tests/test_wasm_persistence.py
@@ -46,6 +46,8 @@ from pathlib import Path
 
 import pytest
 
+import json
+
 STATE_PY = (
     Path(__file__).resolve().parents[2] / "mlsysim" / "mlsysim" / "labs" / "state.py"
 )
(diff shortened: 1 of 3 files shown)


Shashank-Tripathi-07:
Hello Shivtej, Saw your code and did my review on it. Here are the things you should change:

1. Dead branch in a test assertion (mlsysim/tests/test_state.py, test_load_corrupt_file_sets_last_load_error):

`assert "json" in ledger.last_load_error.lower() or "JSON" in ledger.last_load_error`

The .lower() in the first clause already covers everything the second clause checks, since "JSON" in x can never be true when "json" in x.lower() is false. Second clause is unreachable, can just be:

`assert "json" in ledger.last_load_error.lower()`

3. Duplicate imports from the appended test block (mlsysim/tests/test_state.py): the new section (starting at the second module docstring) re-imports pytest and DesignLedger, which are already imported at the top of the file. Not a bug, but worth folding into the existing top-of-file imports to avoid two import blocks in one file.

4. Inconsistent monkeypatch target for Path.home: the existing tests patch state_mod.Path.home (via _ledger_with_home), the new ledger fixture patches pathlib.Path.home directly. Same underlying class object, so both work, but picking one convention would keep the file consistent.

5. import json placement (labs/tests/test_wasm_persistence.py): added between the existing import pytest and STATE_PY = (...), rather than grouped with the stdlib imports above it.

Once you fix this, I'd be happy to merge this into the codebase. Good work till now !

Shashank-Tripathi-07 (approved):

Shashank-Tripathi-07:
lgtm. tagging professor for a review from his end.

profvjreddi:
Thanks @ShivtejG236, and welcome. Integrated into dev.

Your six new tests pass here alongside the existing five. Good first PR to pick, and thanks for the patience on the review.

I applied your commit onto a review branch off current dev rather than merging the PR head, so GitHub will show this as closed instead of merged. Authorship and credit are intact.

@all-contributors please add @ShivtejG236 for bug, code, test in mlsysim

ShivtejG236:
> Thanks @ShivtejG236, and welcome. Integrated into dev.
> 
> Your six new tests pass here alongside the existing five. Good first PR to pick, and thanks for the patience on the review.
> 
> I applied your commit onto a review branch off current dev rather than merging the PR head, so GitHub will show this as closed instead of merged. Authorship and credit are intact.
> 
> @all-contributors please add @ShivtejG236 for bug, code, test in mlsysim
> 

Thank you, Professor!

Was an amazing experience working on this. As my first real OSS contribution, I learnt a lot, especially through the review process

PR #2008, opened 2026-08-11
fix(tinytorch): tito module complete silently skips integration tests for 5 modules

Shashank-Tripathi-07:
## Bug

`tito module complete`'s Step 3 (integration tests) reports "No integration tests for this module" for modules 15 (Quantization), 16 (Compression), 17 (Acceleration), 19 (Benchmarking), and 20 (Capstone), even though real test files exist for all five.

## Root cause

`_run_integration_tests` only looks for `tests/<module>/test_<module>_progressive.py`, falling back to any `test_*_progressive.py` in the module's test directory. Modules 15, 16, 17, 19, and 20 use a different naming convention entirely (`test_<topic>_core.py`, `test_<topic>_integration.py`, no `_progressive` anywhere in the name), so the fallback never matches and the lookup silently reports zero tests instead of surfacing an error.

39 real tests across these 5 modules (about 980 lines of test code) were never executed by the standard student workflow. If a student's own from-scratch implementation had a genuine bug in one of these modules, `tito module complete` could never catch it and would still print "Module Complete!"

## Fix

Widen the fallback: if neither the primary filename nor any `test_*_progressive.py` file exists in the module's test directory, run every `test_*.py` file found there instead of reporting no tests. Modules using the standard naming are unaffected.

## Testing

Verified against a fresh `dev` install with all 20 modules completed:
- Modules 15, 16, 17, 20: previously 0 tests run, now correctly find and run all real tests (8, 8, 6, 9 tests respectively), all passing.
- Module 19: correctly finds and runs 8 tests, 1 fails (`test_multiple_runs_are_consistent`, a timing coefficient-of-variation assertion). Confirmed this is a pre-existing flaky test unrelated to this fix, not previously caught because it was never run; reproduces standalone under load (CV 2.53 vs required < 1.0). Left untouched, out of scope for this fix.
- Module 01 (standard `_progressive` naming): unaffected, still finds and runs its 10 tests exactly as before.

## Test plan

- [ ] `tito module complete 15/16/17/19/20` on a clean checkout, confirm Step 3 now runs real tests instead of reporting none
- [ ] `tito module complete 01` (or any early module), confirm no change in behavior

### DIFF
--- a/tinytorch/tests/19_benchmarking/test_benchmark_core.py
+++ b/tinytorch/tests/19_benchmarking/test_benchmark_core.py
@@ -22,6 +22,7 @@ WHAT WE TEST:
 import pytest
 import numpy as np
 rng = np.random.default_rng(7)
+import statistics
 import sys
 from pathlib import Path
 
@@ -127,6 +128,15 @@ class TestBenchmarkMetrics:
 
         WHY: Benchmarks should be reproducible. Large variance
         means we can't trust the measurements.
+
+        NOTE: raw coefficient of variation (std/mean) on wall-clock latency
+        of a sub-millisecond forward pass is dominated by OS scheduling
+        jitter on shared/virtualized CI runners: a single delayed sample
+        can spike std enough to fail a mean-based check even though every
+        other measurement was tight. Using the median and median absolute
+        deviation instead is standard practice for exactly this kind of
+        noisy, outlier-prone timing data, and still fails on genuinely
+        inconsistent (not just occasionally-delayed) measurements.
         """
         class SimpleModel:
             def __init__(self):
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:
Addressed: added a commit that fixes \`test_multiple_runs_are_consistent\` (module 19), which this PR causes to run for the first time and which was apparently timing out Windows CI.

Root cause: the test measures wall-clock latency of a sub-millisecond forward pass and asserts a raw coefficient of variation (std/mean) under 100%. On a shared/loaded CI runner, one OS-scheduling-delayed sample is enough to spike std and fail the assertion, even when every other measurement was tight.

Fix: replaced the std/mean check with median absolute deviation relative to the median, a standard robust-statistics substitute for outlier-prone timing data. Bumped measurement_runs from 10 to 20 for a more stable statistic (negligible cost given the sub-millisecond operation). No production code touched, only the test's assertion strategy.

Verified 5 consecutive runs all pass in ~0.13-0.17s each on a heavily loaded local machine, where the original version would occasionally fail.

PR #2009, opened 2026-08-11
fix(tinytorch): tito milestone status falsely claims full mastery

Shashank-Tripathi-07:
## Bug

`tito milestone status` shows "Capabilities Unlocked: 6/6, Overall Progress: 100%" and a "FULL MASTERY ACHIEVED / QUEST COMPLETE! / You are now an ML Systems Engineer!" banner the moment all 20 coding modules are done, before a student has run a single milestone script. Every individual milestone panel correctly shows the unlocked icon (not achieved), but the summary banner ignores that distinction.

## Root cause

`get_milestone_status()` only ever computed `is_unlocked` (prerequisite modules complete, ready to run) and used that for both the per-milestone icon and the completion banner. The progress file already separately tracks `completed_milestones` (set only when `tito milestone run` actually succeeds, in `_handle_run_command`), but nothing read it for status display.

## Fix

- `get_milestone_status()` now also computes `is_completed` per milestone and `status["total_completed"]`, reading the existing `completed_milestones` list.
- The per-milestone status indicator now shows a distinct achieved state (checkmark) ahead of the unlocked state, so unlocked-but-not-run and actually-achieved are visually different.
- The status header shows both "Capabilities Unlocked" and "Milestones Achieved" counts.
- "FULL MASTERY ACHIEVED" now requires `total_completed == total milestones`, not `total_unlocked`. Added a separate, honest "All Milestones Ready" panel for the unlocked-but-none-run case, so students still get positive feedback without a false claim.

## Testing

Verified against a fresh `dev` install with all 20 modules completed (prerequisites for all 6 milestones met) but zero milestones run:
- Before: "Capabilities Unlocked: 6/6" and "FULL MASTERY ACHIEVED" displayed immediately.
- After: "Capabilities Unlocked: 6/6, Milestones Achieved: 0/6" and "All Milestones Ready" (not full mastery).
- After running `tito milestone run 01`: "Milestones Achieved: 1/6", Milestone 01's panel shows the achieved icon, other five still show unlocked.

## Test plan

- [ ] Complete all 20 modules without running any milestones, run `tito milestone status`, confirm it does not claim full mastery
- [ ] Run `tito milestone run 01`, confirm status shows 1/6 achieved and milestone 01's icon changes
- [ ] Run all 6 milestones, confirm "FULL MASTERY ACHIEVED" now displays correctly

### DIFF
--- a/tinytorch/tito/commands/milestone.py
+++ b/tinytorch/tito/commands/milestone.py
@@ -672,13 +672,19 @@ class MilestoneCommand(BaseCommand):
         milestone_system = MilestoneSystem(self.config)
         status = milestone_system.get_milestone_status()
 
-        # Show header with overall progress
+        # Show header with overall progress. Note: status['overall_progress']
+        # is unlock-based (it also drives the timeline progress bar elsewhere,
+        # where that's the correct meaning), so it isn't used here -- showing
+        # it next to "Milestones Achieved" would be contradictory (e.g. "0/6
+        # achieved" beside "100%"). This header's percentage is achievement-
+        # based instead, to actually match the achieved count shown above it.
         total_milestones = len(milestone_system.MILESTONES)
+        achievement_progress = (status['total_completed'] / total_milestones) * 100 if total_milestones > 0 else 0
         console.print(Panel(
             f"[bold cyan]🎮 TinyTorch Milestone Progress[/bold cyan]\n\n"
             f"[bold]Capabilities Unlocked:[/bold] {status['total_unlocked']}/{total_milestones} milestones\n"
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
Good catch. Addressed: added a commit that computes a separate, achievement-based percentage for this status header instead of reusing \`status['overall_progress']\` (which is unlock-based, and stays that way since the timeline view elsewhere still needs unlock-based progress for its bar -- changing its meaning globally would've just moved the contradiction there instead).

Verified with 3/6 milestones actually run: header now shows "Milestones Achieved: 3/6 milestones" and "Overall Progress: 50%", consistent instead of contradictory.

PR #2011, opened 2026-08-11
fix(tinytorch): tito module start/resume/view leak untracked Jupyter Lab processes

Shashank-Tripathi-07:
## Bug

Every `tito module start` (without `--no-jupyter`), `resume`, or `view` call launches a brand new `jupyter lab` `subprocess.Popen` with no PID tracking, no cleanup, and no check for whether one is already running. Over a normal working session where a student opens several modules, these accumulate indefinitely: confirmed 45 leaked processes after routine module-by-module testing, and separately confirmed that a build-up of ~38 live kernels was enough to make the Jupyter Lab UI itself sluggish (menu clicks timing out).

## Fix

Track the launched Jupyter Lab process's PID in `.tito/jupyter.pid`. Before launching a new one, check whether that PID is still alive and is actually still a jupyter process (guards against a recycled PID being mistaken for a live server). If so, reuse it: tell the student which notebook to open in their existing tab instead of spawning another server. Only launch a new process if no tracked server is currently running.

## Testing

Verified against a fresh `dev` install with all 20 modules completed:
- First `tito module view 01`: launches Jupyter Lab as before, records its PID.
- Second `tito module view 01` (simulating a student opening another module in the same session): before the fix, this would spawn a second full Jupyter Lab server. After the fix, detects the existing PID, prints "Jupyter Lab is already running (pid N)" with the notebook to open, and spawns nothing. Confirmed via process listing that the `jupyter.exe` count stayed at 1 across both calls.

## Test plan

- [ ] `tito module start N` (without `--no-jupyter`), confirm Jupyter Lab launches and `.tito/jupyter.pid` is written
- [ ] `tito module view M` for a different already-started module while the first server is still running, confirm it reuses the existing server instead of launching a second one
- [ ] Manually kill the tracked process, run another `tito module view`, confirm it correctly detects the server is gone and launches a fresh one

### DIFF
--- a/tinytorch/tito/commands/module/workflow.py
+++ b/tinytorch/tito/commands/module/workflow.py
@@ -501,6 +501,40 @@ class ModuleWorkflowCommand(BaseCommand):
 
         return self._open_jupyter(module_name)
 
+    def _jupyter_pid_file(self) -> Path:
+        return self.config.project_root / ".tito" / "jupyter.pid"
+
+    def _running_jupyter_pid(self) -> Optional[int]:
+        """Return the PID of a tito-launched Jupyter Lab server still running, if any."""
+        pid_file = self._jupyter_pid_file()
+        if not pid_file.exists():
+            return None
+        try:
+            pid = int(pid_file.read_text().strip())
+        except (ValueError, OSError):
+            return None
+
+        try:
+            import psutil
+            if not psutil.pid_exists(pid):
+                return None
+            proc = psutil.Process(pid)
+            cmdline = " ".join(proc.cmdline()).lower()
+            if "jupyter" not in cmdline:
+                # PID was recycled by an unrelated process since we last launched
+                return None
+            return pid
+        except ImportError:
+            # No psutil available - fall back to os-level existence check only,
+            # which can't verify it's still actually a Jupyter process.
+            try:
+                os.kill(pid, 0)
+                return pid
+            except OSError:
+                return None
+        except Exception:
+            return None
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
Addressed two points:

**Windows os.kill(pid, 0) bug (real problem, not a nit):** confirmed and fixed. Removed the unsafe fallback entirely -- on ImportError, \`_running_jupyter_pid\` now just returns \`None\` (treated the same as "can't verify") instead of attempting an os.kill probe. psutil is already a required project dependency (requirements.txt, tito setup's own package list), so this path should only be hit in a broken environment, and even then the worst case is now a duplicate server spawn, never an accidental kill. Verified in isolation with psutil's import faked to raise ImportError: the function returns None and the target process (verified against my own running PID) stays alive throughout.

**Duplicate hunks with #2010:** rebased this branch directly onto \`dev\` (via \`git rebase --onto dev <old-base> <branch>\`), dropping the #2010 commit that had ended up as an ancestor here. The diff against \`dev\` is now clean: only the two commits that actually belong to this PR (the leak fix + the os.kill fix), 55 lines total, no overlap with #2010. Force-pushed the rewritten branch.

No merge-order dependency remains for this PR specifically, though #2010 should still merge before or independently of this one since they touch the same file's nearby regions.

Shashank-Tripathi-07:
Superseded by #2092, which combines this with #2026 since both touch the same file (tito/commands/module/workflow.py). Closing to avoid two PRs churning the same file back to back.

PR #2015, opened 2026-08-11
fix(tinytorch): tito package reset is unreachable

Shashank-Tripathi-07:
## Bug

`tito package reset` (with or without piped input) does nothing but print a generic help panel every time, regardless of input. The documented example in `PackageCommand`'s own help text, `tito package reset --force`, silently does nothing at all.

## Root cause

`ResetCommand` was written as if it would be registered as its own top-level `tito reset <subcommand>` command: its `name` property returns `"reset"`, and its own help text says `tito reset progress --backup`. It is actually only ever mounted at `tito package reset`. By the time argparse reaches this class's own `add_subparsers()` call, both the `package` and `reset` tokens are already consumed by the parent parser, so a bare `tito package reset` never has a SUBCOMMAND token left to match, and `args.reset_command` is always unset. The only way to actually trigger a reset was the undocumented double-noun `tito package reset package --force`.

## Fix

- Added a `--force` argument directly on the reset parser itself (not just on each sub-subparser), so it's available for a bare invocation.
- Bare `tito package reset [--force]` (no SUBCOMMAND) now defaults to resetting the package, matching `PackageCommand`'s own documented example and this command's stated top-level purpose ("Reset tinytorch package to clean state"). It still shows the other reset types (`all`/`progress`/`milestones`/`config`) for discoverability, then proceeds with the actual package reset (still gated behind its own confirmation prompt unless `--force` is passed).
- `tito package reset package --force` (the old double-noun form) still works unchanged.

## Testing

Verified against a fresh `dev` install with all 20 modules exported:
- `tito package reset`, answering "n": shows the confirmation prompt and correctly cancels. Before the fix, this printed only the self-referential help panel and never reached a confirmation at all.
- `tito package reset --force`: correctly removes all 20 generated package files, matching the documented example exactly.
- `tito package reset package --force`: still works as before.

## Test plan

- [ ] `tito package reset`, confirm it shows a real confirmation prompt (not just a help panel)
- [ ] `tito package reset --force`, confirm it actually removes exported package files
- [ ] `tito package reset package --force`, confirm the old form still works

### DIFF
--- a/tinytorch/tito/commands/package/reset.py
+++ b/tinytorch/tito/commands/package/reset.py
@@ -22,6 +22,11 @@ class ResetCommand(BaseCommand):
         return "Reset package files or user progress data"
 
     def add_arguments(self, parser: ArgumentParser) -> None:
+        # Lets bare `tito package reset [--force]` (with no SUBCOMMAND token)
+        # reset the package directly, matching the "tito package reset --force"
+        # example in PackageCommand's own help text.
+        parser.add_argument("--force", action="store_true", help="Skip confirmation prompt (resets the package)")
+
         subparsers = parser.add_subparsers(
             dest='reset_command',
             help='Reset subcommands',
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thanks @Shashank-Tripathi-07. Integrated into dev, CI green.

One repair on the way in. Each subparser already defines its own `--force`, and argparse writes a subparser's default over a shared parent dest, so `tito package reset --force package` silently dropped the flag and then blocked on `input()`. Under a script or in CI that is an EOFError, which is the opposite of what someone passing `--force` wants:

```
$ tito package reset --force package < /dev/null
EOFError: EOF when reading a line
```

Gave the parent flag its own dest and OR the two in `run()`, so `--force` works on either side of the SUBCOMMAND token. `run()` also has to set `args.force` itself, since the bare form parses no subparser and the attribute otherwise doesn't exist.

PR #2019, opened 2026-08-11
fix(tinytorch): tito milestone info duplicates the year in its title

Shashank-Tripathi-07:
## Bug

`tito milestone info 01` renders the title as "Perceptron (1958) (1958)".

## Root cause

`_handle_info_command` builds the title as `"{milestone['name']} ({milestone['year']})"`, but `milestone['name']` already includes the year (e.g. "Perceptron (1958)") in `MILESTONE_SCRIPTS`, so it gets appended a second time.

## Fix

Drop the redundant `({milestone['year']})` suffix; `milestone['name']` already carries it.

## Testing

Verified: `tito milestone info 01` now shows "Perceptron (1958)" once, not twice.

## Test plan

- [ ] `tito milestone info 01` (or any milestone), confirm the year appears only once in the title

### DIFF
--- a/tinytorch/tito/commands/milestone.py
+++ b/tinytorch/tito/commands/milestone.py
@@ -1362,7 +1362,7 @@ class MilestoneCommand(BaseCommand):
             if next_id in MILESTONE_SCRIPTS:
                 next_milestone = MILESTONE_SCRIPTS[next_id]
                 console.print(f"\n[bold yellow]🎯 What's Next:[/bold yellow]")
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
Added a commit fixing the identical pattern at the "What's Next" preview line (was showing "Milestone 04: CNN Revolution (1998) (1998)"). Same root cause, same fix.

Also grepped the whole file for the pattern (\`['name'].*({.*['year']}\`) to confirm no other instances remain. this was the only other one.

PR #2029, opened 2026-08-14
fix: restore broken star history chart in README

OctoBored:
The star history chart on the README is currently broken and no longer renders because the upstream service is affected by GitHub stargazer API restrictions.

This change points the chart (and its localized versions) to a working mirror, so the chart displays again for all visitors. No other content is modified.


Shashank-Tripathi-07:
nice, lgtm.

Shashank-Tripathi-07 (approved):

profvjreddi:
Thanks @OctoBored! I just checked and the chart is rendering again from star-history.com, so I'm going to keep the README pointed at the original service rather than a mirror. Appreciate you keeping an eye on it.

PR #2031, opened 2026-08-14
test(tinytorch): regression coverage for core tensor/layers/autograd/optimizer gaps

Shashank-Tripathi-07:
## Summary

Adds regression tests for core tensor/layers/autograd/optimizer gaps found by an MC/DC-focused audit of the module test suites. Independent of the other companion PRs in this series, each covers a different module group and can be reviewed/merged in any order.

- `Tensor([])` empty-list construction
- 1D-by-2D matmul shape handling
- scalar-form `reshape(n)` and invalid multi-dim/non-divisible `reshape(-1, ...)` cases
- `transpose()` with only one of dim0/dim1 specified
- `Sequential(layer1, layer2)` multi-positional-arg construction (previously only tested via a locally shadowed mock class, not the real class)
- `Dropout`'s bounds validation and training-gate logic (previously zero coverage in the standalone pytest suite)
- `sum(axis=...)` backward gradient correctness
- base `Function.apply()`'s `NotImplementedError`
- `backward()` on a non-scalar tensor without a gradient argument
- SGD's momentum checkpoint API: `has_momentum()`, `get_momentum_state()`, `set_momentum_state()`

## Test plan
- [x] `pytest tests/01_tensor tests/03_layers tests/06_autograd tests/07_optimizers -q` passes locally, including all new cases

### DIFF
--- a/tinytorch/tests/01_tensor/test_tensor_core.py
+++ b/tinytorch/tests/01_tensor/test_tensor_core.py
@@ -568,6 +568,42 @@ class TestTensorPyTorchCompat:
         assert np.array_equal(stacked.data[0], t1.data)
         assert np.array_equal(stacked.data[1], t2.data)
 
+    def test_tensor_from_empty_list(self):
+        """Tensor([]) constructs successfully with shape (0,)."""
+        t = Tensor([])
+        assert t.shape == (0,)
+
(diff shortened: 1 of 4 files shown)


Shashank-Tripathi-07:

PR #2032, opened 2026-08-14
test(tinytorch): regression coverage for dataloader/training checkpoint gaps

Shashank-Tripathi-07:
## Summary

Adds regression tests for dataloader and training checkpoint gaps found by an MC/DC-focused audit of the module test suites. Independent of the other companion PRs in this series, each covers a different module group and can be reviewed/merged in any order.

- `TensorDataset()` with zero tensors raises correctly
- `RandomHorizontalFlip`'s probability-bounds validation (both out-of-range directions)
- `Trainer`'s momentum checkpoint save/restore round trip
- `Trainer`'s scheduler-state checkpoint round trip
- Gradient accumulation's trailing partial-batch flush when batch count doesn't evenly divide `accumulation_steps`

## Test plan
- [x] `pytest tests/05_dataloader tests/08_training -q` passes locally, including all new cases

### DIFF
--- a/tinytorch/tests/05_dataloader/test_dataloader_core.py
+++ b/tinytorch/tests/05_dataloader/test_dataloader_core.py
@@ -25,7 +25,7 @@ from pathlib import Path
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
 from tinytorch.core.tensor import Tensor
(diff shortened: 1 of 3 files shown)


Shashank-Tripathi-07:

PR #2035, opened 2026-08-14
test(tinytorch): regression coverage for milestone grading/gating logic

Shashank-Tripathi-07:
## Summary

Adds regression tests for the milestone system's grading/gating logic, found by an MC/DC-focused audit to be correct but previously exercised only by slow end-to-end runs that couldn't isolate individual conditions. Independent of the other companion PRs in this series, each covers a different module group and can be reviewed/merged in any order.

- The `can_unlock` gate (`MilestoneSystem`)
- `_module_progress_to_int`'s four input-type paths
- `_validate_required_exports`'s `ImportError` branch
- `_handle_run_command`'s non-interactive continue-on-failure abort path
- Milestone 05's `passed1`/`passed2`/`passed3` grading logic in `print_final_results`

## Test plan
- [x] `pytest tests/cli/test_release_regressions.py tests/milestones -q` passes locally, including all new cases

### DIFF
--- a/tinytorch/tests/cli/test_release_regressions.py
+++ b/tinytorch/tests/cli/test_release_regressions.py
@@ -4,11 +4,13 @@ Release regression tests for student-facing CLI and API correctness.
 
 import importlib.util
 import io
+import json
 import os
 import subprocess
 import sys
 from argparse import Namespace
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2036, opened 2026-08-15
fix(tinytorch): TensorDataset rejects valid negative indices

Shashank-Tripathi-07:
## Summary

`TensorDataset.__getitem__` treated any negative index as out of range, so `dataset[-1]` raised `IndexError` instead of returning the last sample. Every other indexable Python container (list, tuple, numpy array) supports negative indexing as counting from the end, so this broke a standard, reasonably expected usage pattern.

Normalizes negative indices the same way Python does before bounds checking, so `dataset[-1]` returns the last item, and an index still too far negative (beyond `-len(dataset)`) continues to raise `IndexError`.

## Test plan
- [x] `pytest tests/05_dataloader -q` passes locally, including the new negative-indexing tests
- [x] Confirmed the bug reproduces on `dev` before the fix, and no longer reproduces after

### DIFF
--- a/tinytorch/src/05_dataloader/05_dataloader.py
+++ b/tinytorch/src/05_dataloader/05_dataloader.py
@@ -430,8 +430,12 @@ class TensorDataset(Dataset):
         - Use generator expression with tuple() for clean syntax
         """
         ### BEGIN SOLUTION
+        original_idx = idx
+        if idx < 0:
+            idx += len(self)
+
         if idx >= len(self) or idx < 0:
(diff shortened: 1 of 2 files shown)


profvjreddi:

One thing changed on the way in: I extended the APPROACH and HINTS above the solution to cover the negative-index normalization. `tito module test 05` runs your new tests against a learner's own implementation, and the spec still told them to check `idx < len(self)`, so they'd have failed on dataset[-1] through no fault of their own.

Shashank-Tripathi-07:
Yessir, the PR has this issue and it's great you spot it and corrected it. Thanks for the help 😁

PR #2037, opened 2026-08-15
fix(tinytorch): Optimizer.__init__ silently discarded a pre-existing param.grad

Shashank-Tripathi-07:
## Summary

`Optimizer.__init__` unconditionally set `param.grad = None` for every parameter, so a gradient a caller computed and assigned before constructing the optimizer was silently wiped with no warning.

Only sets `grad` when the attribute isn't present yet, preserving any gradient a caller already assigned while still guaranteeing every param has a `grad` attribute for tensors created before autograd was enabled.

## Test plan
- [x] `pytest tests/07_optimizers -q` passes locally, including the new tests covering both a pre-existing grad and a param with no grad attribute yet
- [x] Confirmed the bug reproduces on `dev` before the fix, and no longer reproduces after

### DIFF
--- a/tinytorch/src/07_optimizers/07_optimizers.py
+++ b/tinytorch/src/07_optimizers/07_optimizers.py
@@ -257,11 +257,14 @@ class Optimizer:
         # Store parameters
         self.params = params
 
-        # Ensure parameters participate in autograd once it is enabled
+        # Ensure parameters participate in autograd once it is enabled.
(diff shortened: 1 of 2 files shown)


profvjreddi:

Same adjustment as #2036: the APPROACH now says construction must not clear a grad the parameter is already carrying, so what a learner is asked to build matches what your test checks.

PR #2038, opened 2026-08-15
fix(tinytorch): _calculate_improvements silently returned {} on mismatched metric keys

Shashank-Tripathi-07:
## Summary

`_calculate_improvements` only computes a metric's improvement when the key is present in both `base_metrics` and `opt_metrics`, which is correct for legitimate partial overlap. But when the two dicts share no keys at all, it silently returned an empty dict with no signal that the inputs were likely mismatched or malformed.

Adds a warning when the result is empty despite both inputs being non-empty, while leaving the actual computation and partial-overlap behavior unchanged.

## Test plan
- [x] `pytest tests/19_benchmarking/test_benchmark_core.py -q` passes locally, including tests for matching keys, fully mismatched keys, partial overlap, and empty inputs
- [x] Confirmed the silent-empty-dict behavior reproduces on `dev` before the fix

### DIFF
--- a/tinytorch/src/19_benchmarking/19_benchmarking.py
+++ b/tinytorch/src/19_benchmarking/19_benchmarking.py
@@ -3326,6 +3326,15 @@ def _calculate_improvements(base_metrics: Dict[str, float], opt_metrics: Dict[st
         else:
             improvements['accuracy_retention'] = 1.0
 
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2039, opened 2026-08-15
fix(tinytorch): validate_submission_schema was never exported to the package

Shashank-Tripathi-07:
## Summary

`validate_submission_schema` was missing the `#| export` marker every other public function in this module has, so it never made it into the built `tinytorch` package (`tinytorch/olympics.py`). The capstone's only submission validation function was unreachable from anywhere outside the source notebook, including pytest, meaning its negative-path behavior (rejecting malformed submissions) had never actually been exercised by CI.

Adds the export marker so it ships in the package, and adds direct tests covering both the valid-submission path and the negative paths (missing required field, out-of-range accuracy in both directions).

## Test plan
- [x] `pytest tests/20_capstone/test_capstone_core.py -q` passes locally, including new tests importing the function from `tinytorch.olympics` and exercising both valid and invalid submissions
- [x] Confirmed the function was unreachable via `from tinytorch.olympics import validate_submission_schema` on `dev` before the fix

### DIFF
--- a/tinytorch/src/20_capstone/20_capstone.py
+++ b/tinytorch/src/20_capstone/20_capstone.py
@@ -1423,6 +1423,7 @@ This test validates submissions conform to the required schema.
 """
 
 # %% nbgrader={"grade": true, "grade_id": "test-submission-schema", "locked": true, "points": 10}
+#| export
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2040, opened 2026-08-15
fix(tinytorch): CrossEntropyLoss raised a raw numpy IndexError on bad targets

Shashank-Tripathi-07:
## Summary

`CrossEntropyLoss.forward` used target indices directly in numpy fancy indexing with no bounds check, so an out-of-range or negative target class index surfaced as a raw `"index N is out of bounds"` `IndexError` instead of a clear domain-specific error naming the actual problem (an invalid target index) and the valid range.

Module 06's `enable_autograd()` installs a second, separate implementation of cross-entropy forward (`tracked_ce_forward`) to add gradient tracking, which duplicated the same unchecked indexing. Both copies needed the same validation, otherwise the fix would silently stop applying the moment autograd is enabled, which is the normal case in any real training loop.

Both now raise a clear `ValueError` naming the invalid index and the valid class range before attempting the indexing.

## Test plan
- [x] `pytest tests/04_losses -q` passes locally, including new tests for an out-of-range target, a negative target, and valid targets
- [x] `pytest tests/06_autograd -q` still passes (49 passed, 18 skipped), confirming the duplicate autograd-path fix doesn't regress gradient tracking
- [x] Confirmed the raw `IndexError` reproduces on `dev` before the fix, both with and without `enable_autograd()` called

### DIFF
--- a/tinytorch/src/04_losses/04_losses.py
+++ b/tinytorch/src/04_losses/04_losses.py
@@ -634,7 +634,16 @@ class CrossEntropyLoss:
 
         # Step 2: Select log-probabilities for correct classes
         batch_size = logits.shape[0]
+        num_classes = logits.shape[-1]
         target_indices = targets.data.astype(int)
+
+        out_of_range = (target_indices < 0) | (target_indices >= num_classes)
+        if np.any(out_of_range):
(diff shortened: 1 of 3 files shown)


Shashank-Tripathi-07:

PR #2041, opened 2026-08-15
fix(tinytorch): test_sequential_layer_composition tested a shadowed mock, not the real class

Shashank-Tripathi-07:
## Summary

`test_sequential_layer_composition` defined a local `class Sequential(Layer)` that shadowed the `tinytorch.core.layers.Sequential` import, so the test never actually exercised the real container class despite its name and docstring both claiming to test `Sequential` composition.

Removes the local mock and imports the real `Sequential` instead. The real class has a slightly different constructor (`Sequential(*layers)` supporting both `Sequential(l1, l2)` and `Sequential([l1, l2])`, rather than the mock's `Sequential(layers)`-only signature), but the existing call site here already used the list form, so no test logic changed.

## Test plan
- [x] `pytest tests/03_layers/test_layers_core.py -q` passes locally, now genuinely exercising `tinytorch.core.layers.Sequential`

### DIFF
--- a/tinytorch/tests/03_layers/test_layers_core.py
+++ b/tinytorch/tests/03_layers/test_layers_core.py
@@ -28,7 +28,7 @@ from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
-from tinytorch.core.layers import Layer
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2042, opened 2026-08-15
fix(tinytorch): test_10_tokenization_progressive.py never actually tested tokenization

Shashank-Tripathi-07:
## Summary

The file's docstring and every class in it claimed to test Module 10 (Tokenization) plus the prior stack, but zero tests anywhere in the file imported or exercised `tinytorch.core.tokenization`, every class was actually re-testing modules 01-09 (mostly training). A file appearing to gate module 10 in CI was providing no real signal for it.

Adds a `TestModule10TokenizationCore` class covering `CharTokenizer` and `BPETokenizer` encode/decode, and updates the docstring to describe what the file actually covers. The existing prior-stack regression checks are left in place, they're legitimate, just previously mislabeled as being module 10's own coverage.

## Test plan
- [x] `pytest tests/10_tokenization -q` passes locally (22 passed), including the new module-10-specific tests


Shashank-Tripathi-07:

PR #2043, opened 2026-08-15
fix(tinytorch): MLPerf's unknown-benchmark-name check was never run by CI

Shashank-Tripathi-07:
## Summary

`MLPerf.run_standard_benchmark`'s `ValueError` for an unregistered benchmark name was already covered by an inline `test_unit_mlperf_run()` check in the source notebook, but that check only runs manually via `if __name__ == "__main__":` and pytest's `testpaths` is scoped to `tests/`, so it was never actually collected or run by CI.

Adds the equivalent check as a real pytest test so this validation path has genuine CI coverage instead of a check that looked present but silently never ran.

## Test plan
- [x] `pytest tests/19_benchmarking/test_benchmark_core.py -q` passes locally (7 passed), including the new CI-collected test

### DIFF
--- a/tinytorch/tests/19_benchmarking/test_benchmark_core.py
+++ b/tinytorch/tests/19_benchmarking/test_benchmark_core.py
@@ -167,5 +167,30 @@ class TestBenchmarkMetrics:
                 )
 
 
+class TestMLPerfUnknownBenchmarkName:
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2045, opened 2026-08-15
fix(tinytorch): fix big-picture layout escape, card alignment, and dark-mode table contrast

farhan523:
## Summary

Three rendering fixes for the lower half of the TinyTorch **Big Picture** page (`mlsysbook.ai/tinytorch/big-picture.html`). Split into two commits — the first is an independent structural bug, the second is the design work.

## 1. Stray `</div>` closed `<main>` early

An unbalanced `</div>` sat between the module-flow figure and the "Flexible paths" list, taking the page's div depth to `-1`.

Quarto emits its own `<main class="content">` wrapper, so the orphan closed it early. **Everything from "Flexible paths:" to the end of the page** — the learning-path cards, the outcomes table, and all closing sections — was emitted *after* `</main>`, outside the layout grid's body-content track. Two visible consequences:

- That content rendered edge-to-edge with no page margins.
- Quarto only indexes headings inside `<main>`, so the "On this page" TOC silently dropped every section below the figure.

Both `:::` fences above the tag already close their own containers, so it closed nothing opened in this file. Deleting it restores a balanced tree, the page margins, and the full six-entry TOC.

## 2. Ragged card alignment

Each card laid out its body as `<strong>` labels separated by `<br/>`, so the Best for / Time / Outcome rows started at whatever height the module-range subtitle ended at — one line in some cards, two in others. Replaced with a two-column `<dl>` plus a reserved two-line subtitle, so every card's rows share a baseline.

## 3. Dark mode: colour coding erased, table unreadable

**Cards.** The inline `style="background: #e3f2fd"` fills were caught by the hex-normalisation block in `dark-mode.scss`, which flattens those exact strings to a blanket `#2d2d2d !important`. All four tracks rendered as the same grey with accent borders dropped, so the colour carried no information. Each card now declares one `--path-accent` and derives its fills via `color-mix()`, letting dark mode retint per track. This is the refactor that block's own comment asks for — *"Until every page is refactored to `.who-card` / `.callout-note`"*.

**Table.** `shared/styles/partials/_tables.scss` hardcodes dark-on-light text (`#495057` / `#5a6369`) with no dark counterpart, landing at **~1.9:1** on the `#1a1a1a` body — well under the WCAG AA 4.5:1 minimum.

| Element | Before | After |
|---|---|---|
| Header / row label | 1.9:1 | **15.89:1** |
| Body cells | 1.9:1 | **13.33:1** |
| Caption | 1.9:1 | **11.72:1** |

The caption needed its own rule because Quarto renders it as a sibling `<figcaption>`, outside the `<table>`. Light mode is unchanged at 8.18:1.

Table cells also started flush at the container's left edge while prose started at the content margin; zeroing the outer-edge cell padding puts table text on the same line as body copy.

## Scope

- Overrides live in tinytorch's own stylesheets, **not** the shared partial (consumed by every volume).
- They reuse the existing `[data-bs-theme="dark"]` scoping, so they cover both true dark mode and the hybrid state.
- New `.path-*` classes are used only by `big-picture.qmd`.
- PDF output keeps its original markdown via `content-visible` blocks.

One note for reviewers: the table cell-padding rule is intentionally unscoped, so it applies to all TinyTorch tables. I checked `datasets.html` — no regression, and it picks up the same dark-mode contrast improvement. Happy to scope it to this page instead if you'd prefer.

## Verification

Measured in the rendered DOM (not by eye) in light + dark at 1900 / 1280 / 390px:

- All six section headings inside `<main>`; TOC lists all six.
- Prose, cards, and table share the body-content offset (`576` / `301` / `38.3`), no horizontal overflow.
- All four card `<dl>` elements start at an identical offset from their card top.
- Contrast ratios as tabled above.

### DIFF
--- a/tinytorch/quarto/assets/styles/dark-mode.scss
+++ b/tinytorch/quarto/assets/styles/dark-mode.scss
@@ -267,6 +267,35 @@ div[role="doc-toc"] {
   border: 1px solid $border-color-dark;
 }
 
+// --- Learning-path cards in dark mode ---
+// These used to be inline-styled panels caught by the hex-normalisation
+// block below, which flattened all four tracks to the same #2d2d2d and
+// dropped their accent borders — the colour coding carried no information.
+// Because .path-card derives every fill from --path-accent, dark mode only
+// has to retune the mix ratios: each track keeps its hue at a low enough
+// saturation to sit on the #1a1a1a body.
+.path-card {
+  background: color-mix(in srgb, var(--path-accent) 14%, #232323);
+  border-color: color-mix(in srgb, var(--path-accent) 32%, transparent);
+
+  .path-title { color: #f1f5f9; }
+  .path-route { color: #cbd5e1; }
+  dt { color: #e2e8f0; }
+  dd { color: #cbd5e1; }
+
+  dl {
+    border-top-color: color-mix(in srgb, var(--path-accent) 28%, transparent);
+  }
+}
+
+.path-routes {
+  li {
+    border-left-color: color-mix(in srgb, #{$accent-dark} 45%, transparent);
+  }
+
+  strong { color: #f1f5f9; }
+}
+
 // --- Code block surface in dark mode ---
 // Quarto's default styles `div.sourceCode { background: rgba(233, 236, 239, 0.65) }`
 // — a LIGHT fill at 65% opacity. Over the #1a1a1a body that blends to a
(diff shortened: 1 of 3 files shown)


profvjreddi:
Nice catch on the stray </div>, @farhan523. That one was quietly knocking half the page out of the layout and the TOC. I merged dev in to clear the style.scss conflict, and scoped the table padding rule to this page (the outcomes table now sits in a .path-outcomes div) since unscoped it would hit every TinyTorch table. Looked at it in light and dark at desktop and phone widths and it's good. Will merge once CI finishes.

Issue #2046, opened 2026-08-17
Testing TinyTorch


Shashank-Tripathi-07:
## Summary

Tracking issue for an ongoing, systematic testing effort for TinyTorch, inspired by [SQLite's public testing methodology](https://sqlite.org/testing.html) (100% branch coverage, multiple independent test harnesses, fuzzing, cross-checking against reference implementations, mutation testing).

A `testing_tinytorch.md` doc maps SQLite's testing techniques to TinyTorch-specific equivalents and tracks progress: https://github.com/Shashank-Tripathi-07/cs249r_book/blob/docs-testing-tinytorch-philosophy/tinytorch/testing_tinytorch.md

## Done so far

- **MC/DC (Modified Condition/Decision Coverage) audit** across all 20 `src/` modules plus milestone grading/gating logic. ~37 confirmed coverage gaps got regression tests, 8 confirmed real bugs were found and fixed, each with a dedicated regression test and PR (#2025, #2026, #2031-#2043).

## Planned next

1. **Cross-checking against a reference implementation** (SQLite's SLT tests it against other SQL engines; TinyTorch's equivalent is checking `Tensor` ops against real PyTorch/NumPy on identical random inputs). Highest-value item not yet done, since it's the one class of numerical bug that hand-written expected-value unit tests structurally can't catch.
2. Malformed state-file tests (`.tito/progress.json`, `.tito/milestones.json`, checkpoints, tokenizer vocab).
3. Mutation testing (`mutmut`/`cosmic-ray`) to validate how thorough the MC/DC pass actually was.
4. Fuzz testing, systematic boundary-value sweeps, `pytest-cov` in CI, resource/leak testing, fault-injection testing, and multi-Python-version CI, all detailed in the doc linked above.

Will update this issue as each area is worked through.

Shashank-Tripathi-07:
## Update: PyTorch cross-checking done

Built `tests/cross_check/test_pytorch_reference.py` (#2047), comparing TinyTorch against real PyTorch on identical random inputs for arithmetic ops, matmul, reshape/transpose, reductions, all 5 activations, and all 3 losses, forward values and backward gradients both, across 5 independent random draws per test.

Result: 150/150 pass. TinyTorch's implementations are numerically correct against ground truth, not just internally self-consistent.

[...]

PR #2047, opened 2026-08-17
test(tinytorch): add PyTorch cross-checking suite for tensor/autograd ops

Shashank-Tripathi-07:
## Summary

Part of an ongoing testing effort tracked in #2046. Adds `tests/cross_check/test_pytorch_reference.py`, comparing TinyTorch directly against real PyTorch on identical random inputs, for both forward values and backward gradients.

Every other test file in this repo verifies TinyTorch is internally consistent, hand-picked expected values, or finite-difference gradient checks confirming `backward()` agrees with TinyTorch's own `forward()`. Neither can catch a forward pass that's wrong in a self-consistent way, finite differences will confirm a wrong gradient matches a wrong forward pass just as happily as a correct one. Cross-checking against an independent, real reference implementation closes that gap.

Covered: arithmetic ops with broadcasting, matmul, reshape/transpose, sum/mean/max, all five activations, and all three loss functions. GELU is checked against the sigmoid-approximation formula TinyTorch documents using, not PyTorch's default erf-based GELU, which is a different, intentional approximation, comparing against the wrong reference there would fail by design, not due to a bug.

Every test runs against 5 independent random draws via a parametrized autouse fixture rather than a single fixed seed.

`torch` is a **test-only** dependency, added under a new, separate `cross-check` optional-dependencies group (not folded into `dev`), and the suite skips itself cleanly via `pytest.importorskip` if it isn't installed. It's never imported by the `tinytorch` package itself.

A manual stress test at larger input magnitudes than this suite's default range surfaced a real numerical-stability bug in `tracked_sigmoid_forward`, fixed separately in #2048.

## Test plan
- [x] `pytest tests/cross_check -q` passes locally: 150/150 (30 base cases × 5 reseeded draws)
- [x] `pip install -e ".[cross-check]"` installs torch and the suite runs; without it, the suite skips cleanly rather than failing

### DIFF
--- a/tinytorch/pyproject.toml
+++ b/tinytorch/pyproject.toml
@@ -55,6 +55,14 @@ docs = [
     "ipykernel>=6.0.0",
     "ipywidgets>=8.1.8",
 ]
+cross-check = [
+    # Test-only dependency for tests/cross_check/, never imported by the
+    # tinytorch package itself. Compares TinyTorch's forward/backward
+    # output against real PyTorch on identical inputs. Optional: the
+    # suite skips itself cleanly via pytest.importorskip if this isn't
+    # installed, this extra just lets it actually run.
+    "torch>=2.0",
+]
 
 [tool.nbdev]
 lib_name = "tinytorch"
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2048, opened 2026-08-17
fix(tinytorch): tracked_sigmoid_forward used an unstable formula instead of the original

Shashank-Tripathi-07:
## Summary

Module 06's `enable_autograd()` reimplemented `Sigmoid`'s forward pass as `1/(1+exp(-x))` instead of delegating to the original, numerically stable implementation (which branches on sign so every exponent stays <= 0). `tracked_softmax_forward` and `tracked_gelu_forward` right next to it already delegate to their captured originals correctly, sigmoid was the one inconsistent case, and `_original_sigmoid_forward` was already captured but never used.

For large-magnitude inputs (the common case right after gradients explode during training) this raised a spurious `RuntimeWarning` from `np.exp` overflowing. The values happened to still come out correct due to IEEE754 overflow/underflow semantics, so this was a code-quality and warning-noise bug, not a silent wrong-value bug, but the same architectural mistake (a monkey-patched duplicate implementation silently dropping a safety feature the original had) that caused the `CrossEntropyLoss` `IndexError` bug fixed in #2040.

Found via PyTorch cross-checking (#2047), a manual stress test at larger input magnitudes than that suite's default range surfaced the `RuntimeWarning` that motivated this fix.

## Test plan
- [x] `pytest tests/06_autograd tests/02_activations -q` passes locally (97 passed, 18 skipped)
- [x] Confirmed the `RuntimeWarning` reproduces on `dev` before the fix, at scale=10/50/200/1000, and no longer reproduces after
- [x] New regression test asserts no warning fires (via `warnings.simplefilter("error")`) and confirms correct saturation behavior at extreme magnitude

### DIFF
--- a/tinytorch/src/06_autograd/06_autograd.py
+++ b/tinytorch/src/06_autograd/06_autograd.py
@@ -2847,8 +2847,11 @@ def enable_autograd(quiet=False):
 
         def tracked_sigmoid_forward(self, x):
             """Sigmoid with gradient tracking."""
-            result_data = 1.0 / (1.0 + np.exp(-x.data))
-            result = Tensor(result_data)
+            # Delegate to the original, numerically stable implementation
+            # (same pattern as tracked_softmax_forward/tracked_gelu_forward
+            # below) rather than reimplementing sigmoid with a formula that
+            # overflows np.exp for large negative x.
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2049, opened 2026-08-17
fix(tinytorch): malformed-but-syntactically-valid progress/milestones JSON crashed

Shashank-Tripathi-07:
## Summary

Part of an ongoing testing effort tracked in #2046. Every state-file reader in `tito/commands/milestone.py` already guarded against invalid JSON (truncated, empty, binary garbage) via `except (json.JSONDecodeError, IOError)`, falling back to sensible empty defaults. It didn't guard against JSON that parses successfully but isn't the expected object shape, a bare list, string, or number instead of a dict. A bad merge conflict resolution, a manual edit, or any other process writing the wrong structure to `.tito/progress.json` or `.tito/milestones.json` produced valid JSON that then crashed with a raw `AttributeError` (`'list' object has no attribute 'get'`) at the read site or later downstream, instead of the graceful fallback the surrounding code clearly intends.

Fixes all 5 affected call sites: `_load_completed_module_numbers`, `_is_module_completed`, both copies of `_get_milestone_progress_data` (`MilestoneSystem` and `MilestoneCommand`), and the "what's next" progress read in the milestone run command. Each now validates the loaded data is a dict before using it, treating a wrong-shaped file the same as a decode failure.

Found while building a malformed-state-file test suite as a follow-up to the PyTorch cross-checking work (#2047, #2048).

## Test plan
- [x] `pytest tests/cli/test_release_regressions.py tests/milestones -q` passes locally (40 passed, 13 pre-existing unrelated failures on this Windows environment, confirmed present on `dev` before this change too)
- [x] New `tests/milestones/test_malformed_state_files.py` covers wrong-type JSON for both files across all affected call sites, plus a sanity check that genuinely-invalid JSON still falls back correctly and that valid data still round-trips normally
- [x] Confirmed the `AttributeError` reproduces on `dev` before the fix at each of the 5 sites, and no longer reproduces after

### DIFF
--- a/tinytorch/tests/milestones/test_malformed_state_files.py
+++ b/tinytorch/tests/milestones/test_malformed_state_files.py
@@ -1,0 +1,163 @@
+"""
+Malformed .tito state file tests.
+
+TinyTorch's progress/milestone tracking already had a defensive pattern
+in tito/commands/milestone.py: catch (json.JSONDecodeError, IOError) around
+every state-file read and fall back to sensible empty defaults rather than
+crashing. That pattern correctly handles a file that is not valid JSON at
+all (truncated, empty, binary garbage).
+
+It did not handle a file that IS valid JSON but the wrong shape (a bare
+list or string instead of the expected object), a real corruption mode:
+a bad merge conflict resolution, a manual edit gone wrong, or an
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2050, opened 2026-08-17
fix(tinytorch): Trainer.load_checkpoint had no error handling for corrupted files

Shashank-Tripathi-07:
## Summary

Part of an ongoing testing effort tracked in #2046. `load_checkpoint`'s `pickle.load()` call had zero error handling: an empty file raised a raw `EOFError`, garbage bytes raised `pickle.UnpicklingError` with an unhelpful message (`"invalid load key, 'n'"`), and a checkpoint missing expected keys raised a bare `KeyError`. None of these gave any indication of what actually went wrong or how to fix it, unlike the rest of this codebase's well-designed error paths.

Wraps the unpickling in a try/except that raises a clear `ValueError` for a corrupted/incomplete file (interrupted save, disk full, killed process are the realistic causes), validates the loaded object is actually a dict before using it, and validates the expected top-level keys are present, raising a clear error naming what's missing rather than crashing deep inside attribute restoration.

Found while building a malformed-state-file test suite as a follow-up to the PyTorch cross-checking work (#2047, #2048), following the same "corrupted input must never crash confusingly" principle already applied to `.tito/progress.json` and `.tito/milestones.json` in #2049.

## Test plan
- [x] `pytest tests/08_training -q` passes locally (64 passed, 8 skipped)
- [x] New `TestLoadCheckpointMalformedFile` class covers an empty file, garbage bytes, a valid-pickle-wrong-type file, and a valid dict missing required keys, plus a regression guard confirming normal checkpoint round-trips still work
- [x] Confirmed each raw exception type reproduces on `dev` before the fix, and the clear `ValueError` replaces it after

### DIFF
--- a/tinytorch/src/08_training/08_training.py
+++ b/tinytorch/src/08_training/08_training.py
@@ -1401,7 +1401,33 @@ def trainer_load_checkpoint(self, path: str):
     """
     ### BEGIN SOLUTION
     with open(path, 'rb') as f:
-        checkpoint = pickle.load(f)
+        try:
+            checkpoint = pickle.load(f)
+        except (pickle.UnpicklingError, EOFError) as e:
+            raise ValueError(
+                f"Checkpoint file appears corrupted or incomplete: {path}\n"
+                f"  This can happen if the save was interrupted (crash, disk full,\n"
+                f"  killed process) partway through writing the file.\n"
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2051, opened 2026-08-17
fix(tinytorch): several arithmetic operators bypassed Module 01's own implementation

Shashank-Tripathi-07:
## Summary

Part of an ongoing testing effort tracked in #2046, found via mutation testing (`cosmic-ray`) against `tests/01_tensor`.

`enable_autograd()` monkeypatches several `Tensor` arithmetic methods in ways that made Module 01's own implementation unreachable in the fully-built reference package, split across two related but distinct causes:

**1. Reflected operators and `sum()`** (`__radd__`, `__rsub__`, `__rmul__`, `__rtruediv__`, `.sum()`): these were reimplemented directly inside the monkeypatch instead of delegating back to the Module 01 originals, unlike every sibling method (`__add__`, `__sub__`, `__mul__`, `matmul`, `transpose`, `reshape`, `__getitem__`, `__init__`), which already call their captured `_original_*` reference.

**2. The forward operators' scalar branch** (`__add__`, `__sub__`, `__truediv__`): `tracked_add`/`tracked_sub`/`tracked_div` reassigned `other = Tensor(other)` *before* calling `_original_add`/`_original_sub`/`_original_div`, so those originals' own scalar (non-Tensor) branch never ran, only their Tensor-typed branch did. `tracked_mul` already avoided this (keeps a separate Tensor-typed copy for gradient bookkeeping while passing the original `other` through unchanged), which is why `__mul__` had zero mutation survivors while the other three had 11-23 each.

Since `tinytorch/__init__.py` unconditionally calls `enable_autograd()` on package import, both issues meant real chunks of Module 01's `Tensor` class were dead code in any normal test run, no test in this repo, however thorough, could have observed a bug introduced into them. Confirmed directly both times: hand-mutating the affected code and re-running the full test suite left it green before the fix, and failing after.

## Test plan
- [x] `pytest tests/01_tensor tests/02_activations tests/03_layers tests/04_losses tests/06_autograd tests/07_optimizers tests/08_training tests/09_convolutions tests/10_tokenization tests/11_embeddings tests/12_attention tests/13_transformers -q` passes locally (438 passed, 26 skipped)
- [x] Manually verified values and gradients for all affected paths against hand-computed expected values
- [x] New `test_reflected_operator_gradients` and `test_forward_scalar_branch_gradients` lock in both value and gradient for each
- [x] Confirmed via hand-mutation, twice, that the fixes actually restore testability

### DIFF
--- a/tinytorch/src/06_autograd/06_autograd.py
+++ b/tinytorch/src/06_autograd/06_autograd.py
@@ -2505,19 +2505,24 @@ def enable_autograd(quiet=False):
         # Ensure self has gradient attributes
         _ensure_grad_attrs(self)
 
-        # Convert scalar to Tensor if needed
+        # Keep a Tensor-typed copy for grad bookkeeping, but call the
+        # original with `other` in its original form (matching
+        # tracked_mul's pattern) so _original_add's own scalar branch
+        # stays reachable instead of always seeing a pre-converted Tensor.
         if not isinstance(other, Tensor):
-            other = Tensor(other)
-        _ensure_grad_attrs(other)
+            other_tensor = Tensor(other)
+        else:
+            other_tensor = other
+        _ensure_grad_attrs(other_tensor)
 
         # Call original operation
         result = _original_add(self, other)
         _ensure_grad_attrs(result)
 
         # Track gradient if needed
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2053, opened 2026-08-17
test(tinytorch): add softmax backward coverage for non-uniform gradients

Shashank-Tripathi-07:
Part of #2046.

Mutation testing (cosmic-ray) against `tinytorch/core/autograd.py` found that `SoftmaxBackward.apply()`'s gradient formula could be mutated (subtraction to addition, multiplication, division) without any test noticing.

Root cause: every existing softmax gradient test called `.sum()` on the softmax output directly before `backward()`. That forces `grad_output` to be uniform (all ones), and since softmax rows always sum to 1, the gradient's `sum(grad_output * softmax)` term is then always exactly 1, making `grad_output - sum_term` always 0 regardless of whether the operator is `-`, `+`, `*`, or `/`.

Fix: weight the softmax output by a non-uniform tensor before summing, so `grad_output` at the softmax node is non-trivial and the formula is actually exercised. Verified via hand-mutation: the new test fails against the mutated formula and passes against the correct one.


Shashank-Tripathi-07:

PR #2054, opened 2026-08-17
test(tinytorch): add coverage for broadcast gradient shape reduction

Shashank-Tripathi-07:
Part of #2046.

**Correction (posted after opening):** the PR description below originally claimed `_reduce_broadcast_grad()`'s hand-written unit test had "zero actual CI coverage." That's inaccurate. It's true pytest never collects it, but `tito`'s inline-test harness (`tito module complete` / `tito dev test --inline`, which CI runs) executes `src/06_autograd/06_autograd.py` directly as a script, which does call `test_unit_reduce_broadcast_grad()` via its `if __name__ == "__main__":` block. I verified this by hand-mutating the helper and running the source file directly: the inline test does fail. So this helper already has real regression protection, just through a different harness than pytest.

The second half of this PR is not affected by that correction: `Tensor.backward()`'s own copy of the broadcast-reduction logic (triggered by calling `.backward()` directly with an oversized gradient, rather than through an operator's `apply()`) has no equivalent coverage in any inline test either. That part of this PR still closes a real, previously-untested gap.

Given the first half is now redundant with existing inline-test coverage rather than a new gap, I'd understand if only the `TestTensorBackwardBroadcastReduction` class in the added test file is considered worth keeping; the `TestReduceBroadcastGradHelper` class is still correct and passing, just no longer novel coverage.

---

Original description:

_reduce_broadcast_grad() had a hand-written unit test inside the source notebook (src/06_autograd/06_autograd.py), but that test exports into tinytorch/core/autograd.py as a plain function rather than into tests/, so pytest never collects or runs it.

Tensor.backward() also carries its own copy of this reduction logic, used when an external caller passes a gradient shaped for the broadcasted result rather than the original tensor. Every existing test reaches backward() through the internal grad_fn.apply() -> parent backward() chain, where the operator's own apply() already reduced the gradient first, so this second copy was never exercised either. Calling backward() directly with an oversized gradient, a legitimate and documented use of the public API, exercises it.

Verified via hand-mutation on both AND-to-OR and off-by-one boundary mutations in the singleton-dimension check.

### DIFF
--- a/tinytorch/tests/06_autograd/test_broadcast_grad_reduction.py
+++ b/tinytorch/tests/06_autograd/test_broadcast_grad_reduction.py
@@ -1,0 +1,98 @@
+"""
+Tests for gradient shape reduction after broadcasting.
+
+_reduce_broadcast_grad() has a hand-written "unit test" inside the source
+notebook (src/06_autograd/06_autograd.py), but that test is exported into
+tinytorch/core/autograd.py as a plain function, not into the tests/
+directory, so pytest never collects or runs it. This file gives the helper
+actual, CI-executed coverage.
+
+Tensor.backward() also has its own copy of this reduction logic, applied
+whenever an external caller passes a gradient shaped for the broadcasted
+result rather than the original tensor. Every existing test reaches
+backward() through the internal grad_fn.apply() -> parent.backward() chain,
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2055, opened 2026-08-17
test(tinytorch): add coverage for partial requires_grad on binary ops

Shashank-Tripathi-07:
Part of #2046.

No existing test set requires_grad=True on only one operand of a binary arithmetic op. Adds a parametrized check that the non-flagged operand's .grad stays None and the flagged operand's gradient is still correct, across add/sub/mul/div.

Note: mutation testing had flagged ~24 AND-to-OR mutations on the `isinstance(x, Tensor) and x.requires_grad` guards inside each Backward.apply() as surviving. Investigation showed these are true equivalent mutants: Tensor.backward()'s own propagation loop re-checks `tensor.requires_grad` before ever calling tensor.backward() on a parent, so a looser guard inside apply() only wastes a computation that gets discarded, it can't produce an observable difference in any .grad value. This test does not (and structurally cannot) kill those mutants; it closes a real, separate coverage gap: partial requires_grad had no test at all before this.

### DIFF
--- a/tinytorch/tests/06_autograd/test_gradient_correctness.py
+++ b/tinytorch/tests/06_autograd/test_gradient_correctness.py
@@ -184,6 +184,33 @@ class TestArithmeticGradients:
 
         check_grad(fn, np.array([1.0, -1.0, 3.0]))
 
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2056, opened 2026-08-17
test(tinytorch): add loss backward coverage for grad_output scaling and BCE clip boundary

Shashank-Tripathi-07:
Part of #2046.

Mutation testing found that MSEBackward, BCEBackward, and CrossEntropyBackward's final `grad * grad_output` step could be mutated to `/` without any test noticing, and that BCEBackward's clip bound (`1 - eps`) could be mutated to `1 + eps` or `1 / eps` undetected.

Root cause: every existing loss-gradient test reaches `backward()` through `.sum().backward()`, which passes an implicit `grad_output=1.0`, so multiplying and dividing by 1 are indistinguishable. BCE's existing test also deliberately avoids prediction values near 0 or 1 to keep its finite-difference check numerically stable, so the clip bound was never exercised either.

Fix: added direct-value tests (not finite-difference, since finite differences are unstable right at a clip boundary) that call `backward()` with a non-1.0 gradient and, for BCE, a boundary prediction value, then compare against the hand-computed expected gradient. Verified via hand-mutation on all three affected classes.


Shashank-Tripathi-07:

PR #2057, opened 2026-08-17
test(tinytorch): add coverage for SGD momentum-state checkpointing

Shashank-Tripathi-07:
Part of #2046.

SGD.has_momentum(), get_momentum_state(), and set_momentum_state() exist specifically so training checkpoints (Module 08) can save and restore optimizer momentum without hasattr() checks, the same bug category as the corrupted-checkpoint crash already fixed for Trainer.load_checkpoint. Despite that, none of the three methods had any test, pytest or inline: not the has_momentum()/momentum=0 branches, not a save/restore round-trip, and not the length-mismatch error path.

Confirmed via hand-mutation that the length-mismatch guard currently prevents a raw IndexError crash: removing the length check makes set_momentum_state() crash with `IndexError: list assignment index out of range` instead of raising the documented ValueError, when restoring a checkpoint saved with a different parameter count.

### DIFF
--- a/tinytorch/tests/07_optimizers/test_momentum_state_checkpointing.py
+++ b/tinytorch/tests/07_optimizers/test_momentum_state_checkpointing.py
@@ -1,0 +1,117 @@
+"""
+Tests for SGD's momentum-state checkpointing API.
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2058, opened 2026-08-17
test(tinytorch): add exact-value coverage for Adam/AdamW weight decay

Shashank-Tripathi-07:
Part of #2046.

Weight decay had no exact-value test anywhere, pytest or inline, for either optimizer, only "the parameter changed" or "Adam and AdamW disagree" checks. That's not strong enough: hand-mutating AdamW's decoupled decay step from `(1 - lr * weight_decay)` to `(1 + lr * weight_decay)`, which grows weights instead of shrinking them, passes every existing test including the inline ones.

Adam and AdamW apply decay differently on purpose (Adam mixes it into the gradient before the moment update; AdamW applies it as a separate multiplicative step after, decoupled from the adaptive scaling), which is the entire reason AdamW exists. Added hand-computed exact-value tests for both formulas, plus a check that `weight_decay=0` makes them equivalent. Verified via hand-mutation on both optimizers' decay lines.

### DIFF
--- a/tinytorch/tests/07_optimizers/test_weight_decay_exact_values.py
+++ b/tinytorch/tests/07_optimizers/test_weight_decay_exact_values.py
@@ -1,0 +1,101 @@
+"""
+Exact-value tests for weight decay in Adam and AdamW.
+
+Adam and AdamW apply weight decay in deliberately different ways, and the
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2059, opened 2026-08-17
test(tinytorch): add NaN/Inf safety coverage for LayerNorm constant input

Shashank-Tripathi-07:
Part of #2046.

LayerNorm divides by `sqrt(variance + eps)`, specifically to avoid dividing by zero when a sample's values are identical along the normalized dimension (variance = 0), a real scenario for padding token embeddings or other degenerate inputs. The eps guard already exists in both forward and backward, but nothing tested it: no test exercised a constant-valued input to confirm the result actually stays finite.

Verified via hand-mutation: removing eps from the forward sqrt produces genuine NaN (confirmed via numpy's own RuntimeWarning) and every new test catches it, including the backward-pass gradient check.

Part of the NaN/Inf propagation audit mentioned in the testing plan (untested in 16 of 20 modules per an earlier MC/DC pass).

### DIFF
--- a/tinytorch/tests/13_transformers/test_layernorm_numerical_edge_cases.py
+++ b/tinytorch/tests/13_transformers/test_layernorm_numerical_edge_cases.py
@@ -1,0 +1,76 @@
+"""
+Numerical edge-case tests for LayerNorm.
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2060, opened 2026-08-17
fix(tinytorch): GELUBackward overflows on large-magnitude inputs

Shashank-Tripathi-07:
Part of #2046.

GELUBackward.apply() recomputes its own sigmoid term for the gradient using the naive `1/(1+exp(-z))` formula, which overflows `np.exp` for large `|z|` (`z = 1.702*x` here). The final gradient still happens to saturate to the correct value in this case, but the spurious `RuntimeWarning` pollutes logs and breaks under strict warnings-as-errors configurations, the same class of issue already found and fixed for `tracked_sigmoid_forward`.

Rewrote it to use the same branch-based numerically stable sigmoid pattern already established in `activations.Sigmoid.forward`: each branch keeps its exponent `<= 0`, so large `|x|` never overflows.

Verified via hand-mutation: reverting to the naive formula makes the new test fail with the exact overflow warning.

Found during the NaN/Inf propagation audit mentioned in the testing plan.


Shashank-Tripathi-07:

PR #2061, opened 2026-08-17
test(tinytorch): add NaN/Inf safety coverage for BatchNorm2d constant input

Shashank-Tripathi-07:
Part of #2046.

BatchNorm2d divides by `sqrt(var + eps)`, guarding against a channel with zero variance (identical values across the batch and spatial extent, a real scenario for a constant or degenerate feature map). The eps guard already exists, but BatchNorm2d otherwise had essentially no test coverage at all, nothing exercised the constant-channel case to confirm it actually stays finite.

Verified via hand-mutation: removing eps from the forward sqrt produces genuine NaN and three of the four new tests catch it (the eval-mode test correctly stays green, since `running_var` defaults to 1.0 rather than being computed from the batch).

Part of the NaN/Inf propagation audit mentioned in the testing plan.

### DIFF
--- a/tinytorch/tests/09_convolutions/test_batchnorm_numerical_edge_cases.py
+++ b/tinytorch/tests/09_convolutions/test_batchnorm_numerical_edge_cases.py
@@ -1,0 +1,68 @@
+"""
+Numerical edge-case tests for BatchNorm2d.
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2062, opened 2026-08-17
test(tinytorch): add property-based fuzz testing for Tensor construction and reshape

Shashank-Tripathi-07:
Part of #2046.

Hand-written tests check specific, chosen inputs. Fuzz testing generates many randomized inputs (shapes, values, edge cases) to find the case nobody thought to write a test for, complementing rather than replacing the mutation and MC/DC testing already done in this initiative.

Adds `hypothesis` as an optional test-only dependency (the `fuzz` group, never imported by the tinytorch package itself, skips cleanly via `pytest.importorskip` if not installed) and property tests for two of the plan's suggested fuzz targets:

**Tensor construction and reshape:**
- construction preserves shape/size/values for any array
- reshape to any valid target shape preserves the flattened element order
- `-1` inference matches for any valid partial shape
- flatten-then-restore round-trips exactly

**CharTokenizer encode/decode:**
- encode/decode round-trips exactly when the vocabulary covers the text
- encode() never crashes on arbitrary unicode text (unknown characters map to `<UNK>`)
- decode() never crashes on arbitrary or out-of-range token IDs (a malformed token stream must degrade gracefully, not crash)
- encode()'s output length always matches the input length

Both verified via hand-mutation: an off-by-one in reshape's `-1` inference and a dropped-token bug in encode() are both caught immediately, with hypothesis shrinking each failure down to a minimal reproducing case.

### DIFF
--- a/tinytorch/pyproject.toml
+++ b/tinytorch/pyproject.toml
@@ -48,6 +48,14 @@ dev = [
 visualization = [
     "matplotlib>=3.10.9",  # For plotting in Modules 17, 19, 20
 ]
+fuzz = [
+    # Test-only dependency for property-based fuzz tests, never imported by
+    # the tinytorch package itself. Generates randomized inputs (shapes,
+    # dtypes, edge-case values) to find cases no one thought to hand-write
+    # a test for. Optional: fuzz tests skip themselves cleanly via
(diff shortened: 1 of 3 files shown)


Shashank-Tripathi-07:

PR #2063, opened 2026-08-17
test(tinytorch): add systematic boundary-value coverage

Shashank-Tripathi-07:
Part of #2046.

Adds the boundary-value sweep from the testing plan:

- **DataLoader**: a partial last batch (dataset size not evenly divisible by batch_size), `batch_size=1`, `batch_size` equal to or larger than the dataset size.
- **Attention**: single-token sequences (`seq_len=1`, the first step of autoregressive generation), where the causal mask degenerates to a 1x1 all-ones matrix.
- **Conv2d/MaxPool2d/AvgPool2d**: `kernel_size` equal to the input's spatial size (the "global" convolution/pooling boundary), including the fully degenerate 1x1 input case.

None of these were previously exercised, only evenly-divisible batch sizes and kernel sizes smaller than the input were tested.

Verified via hand-mutation: a floor-division bug in `DataLoader.__len__`, an unscaled attention output, and an off-by-one in Conv2d's output-size formula are all caught by the corresponding new tests.

### DIFF
--- a/tinytorch/tests/05_dataloader/test_dataloader_boundary_values.py
+++ b/tinytorch/tests/05_dataloader/test_dataloader_boundary_values.py
@@ -1,0 +1,93 @@
+"""
+Boundary-value tests for DataLoader batching.
+
+Systematic sweep of batch-size/dataset-size edge cases: a dataset size
+not evenly divisible by batch_size (the last batch is partial), a single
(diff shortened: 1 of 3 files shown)


Shashank-Tripathi-07:

PR #2064, opened 2026-08-17
test(tinytorch): add boundary coverage for log_softmax masking

Shashank-Tripathi-07:
Part of #2046.

log_softmax had zero pytest coverage before this (only an inline test existed). Masking invalid classes with -inf before log_softmax is a real, common pattern (attention padding, invalid-action masking in RL), so both the safe case (some classes masked) and the degenerate case (every class masked) are worth locking in explicitly.

Verified against real PyTorch: a fully-masked row produces NaN in both implementations, this is intentional parity with the reference, not a bug to fix (confirmed via `torch.log_softmax` directly).

Verified via hand-mutation that per-row (not global) max and sum are actually load-bearing for correctness: a naive constant-shift mutation on the max turned out to be mathematically invariant under the log-sum-exp trick (only affects numerical stability, not the final answer), so the regression test specifically targets the sum's axis parameter, which does change the answer when broken.

### DIFF
--- a/tinytorch/tests/04_losses/test_log_softmax_masking.py
+++ b/tinytorch/tests/04_losses/test_log_softmax_masking.py
@@ -1,0 +1,70 @@
+"""
+Boundary tests for log_softmax with masked (-inf) logits.
+
+log_softmax had zero pytest coverage before this file (only an inline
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2065, opened 2026-08-17
fix(tinytorch): milestone JSON reading crashed on non-cp1252 unicode and null/wrong-type completed_modules

Shashank-Tripathi-07:
Note: this branch is built on top of #2049 (unmerged), since it fuzzes the exact function that PR fixed. Until #2049 merges, this diff will show both PRs' changes combined; the changes specific to this PR are the encoding fix, the completed_modules type-safety helper, and the new fuzz/regression tests described below.

Part of #2046.

Property-based fuzz testing (hypothesis, extending the malformed-JSON regression tests from #2049) found two more real bugs in `_load_completed_module_numbers()` and its two call sites reading `progress.json`:

1. `open(progress_file, 'r')` had no explicit encoding, so on Windows it defaults to the system locale (cp1252), not UTF-8. Any `progress.json` containing a Unicode character outside cp1252's range raised a raw `UnicodeDecodeError`, uncaught since the except clause only handled `(json.JSONDecodeError, IOError)`. Fixed by opening with `encoding='utf-8'` explicitly (all read and write sites) and adding `UnicodeDecodeError` to the caught exceptions as defense in depth.

2. `progress_data.get("completed_modules", [])` only substitutes the default when the key is absent, not when its value is present but the wrong type: `{"completed_modules": null}` or `{"completed_modules": true}` (a real corruption mode from a bad merge or partial write) returned `None`/`True` respectively, and iterating that raised a raw `TypeError`. Added a `_get_completed_modules_list()` helper that validates the value is actually a list before using it, applied at all three call sites.

Both verified via hand-mutation (reverting each fix independently reproduces its crash) and locked in with hand-picked regression tests in `test_malformed_state_files.py`, plus a new property-based fuzz test file covering the full space of malformed `progress.json` shapes.

### DIFF
--- a/tinytorch/pyproject.toml
+++ b/tinytorch/pyproject.toml
@@ -48,6 +48,14 @@ dev = [
 visualization = [
     "matplotlib>=3.10.9",  # For plotting in Modules 17, 19, 20
 ]
+fuzz = [
+    # Test-only dependency for property-based fuzz tests, never imported by
+    # the tinytorch package itself. Generates randomized inputs (shapes,
+    # dtypes, edge-case values) to find cases no one thought to write a
+    # test for. Optional: fuzz tests skip themselves cleanly via
+    # pytest.importorskip if this isn't installed.
+    "hypothesis>=6.100",
+]
 docs = [
     "jupyter-book>=2.1.5,<3.0.0",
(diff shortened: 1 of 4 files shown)


Shashank-Tripathi-07:

PR #2066, opened 2026-08-17
test(tinytorch): add boundary coverage for Embedding and PositionalEncoding

Shashank-Tripathi-07:
Part of #2046.

Systematic sweep: `vocab_size=1` (the smallest valid vocabulary), an out-of-range token index (must raise a clean ValueError, not an obscure numpy IndexError from deep inside advanced indexing), a negative index (must not silently wrap around via numpy's negative-indexing semantics), and `embed_dim=1` / `max_seq_len=1` for positional encoding. None of these were previously exercised.

Verified via hand-mutation: an off-by-one in the out-of-range check (`>` instead of `>=`) is caught, and demonstrates the exact failure mode this test guards against: a confusing raw IndexError instead of the documented ValueError.

### DIFF
--- a/tinytorch/tests/11_embeddings/test_embeddings_boundary_values.py
+++ b/tinytorch/tests/11_embeddings/test_embeddings_boundary_values.py
@@ -1,0 +1,83 @@
+"""
+Boundary-value tests for Embedding and PositionalEncoding.
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2067, opened 2026-08-17
fix(tinytorch): tito module start/resume spawns duplicate Jupyter servers

Shashank-Tripathi-07:
Part of #2046.

`_open_jupyter()` (used by both `tito module start` and `tito module resume`) had no check for an already-running Jupyter Lab server before spawning a new one via `subprocess.Popen`. Calling resume more than once, e.g. after forgetting a previous session was still open, silently piled up untracked Jupyter server processes with no way to clean them up short of finding and killing them manually. This is the resource-leak class of bug flagged in the testing plan.

Added `_find_running_jupyter_lab_pids()`, a psutil-based scan for processes whose command line contains both "jupyter" and "lab". If one is already running, `_open_jupyter()` now reuses it (prints a message and returns) instead of spawning a duplicate.

Matches the existing codebase convention of treating psutil as optional (try/except ImportError, already used in `tito/commands/system/info.py`): if psutil isn't installed or the scan fails for any reason, the check returns `None` and launch proceeds as before rather than blocking a legitimate start.

Tested by mocking psutil directly (spawning real Jupyter processes in CI would be slow and environment-fragile) plus one live check against the real psutil integration path against actual running system processes.

### DIFF
--- a/tinytorch/tests/cli/test_jupyter_duplicate_detection.py
+++ b/tinytorch/tests/cli/test_jupyter_duplicate_detection.py
@@ -1,0 +1,117 @@
+"""
+Tests for _find_running_jupyter_lab_pids(), the duplicate-process guard
+for `tito module start`/`resume`.
+
+Repeated `tito module resume` calls previously spawned a brand new
+Jupyter Lab server process every time, with nothing tracking or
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2068, opened 2026-08-17
fix(tinytorch): Trainer.save_checkpoint not atomic, crash mid-write corrupts checkpoint

Shashank-Tripathi-07:
Part of #2046.

`save_checkpoint` wrote directly to the target path: `open(path, 'wb')` then `pickle.dump(checkpoint, f)`. If the process is killed, the disk fills up, or pickling fails partway through (all realistic during a long training run), the target file is left partially written, silently destroying whatever good checkpoint was there before. Unlike a read-time corruption (already handled gracefully by `Trainer.load_checkpoint`'s error handling), this is unrecoverable: the previous valid checkpoint is gone, along with however many hours of training it represented.

Fixed with the standard write-temp-then-atomic-replace pattern: `pickle.dump` to a `.tmp` file, then `os.replace()` into the real path. `os.replace()` is atomic on both POSIX and Windows, so the target is always either the old complete checkpoint or the new one, never a partial write. The temp file is cleaned up on any failure.

Verified via hand-mutation: reverting to the direct-write pattern reproduces both failure modes these tests catch, a corrupted target file and a leftover `.tmp` file.

Found during the anomaly/fault-injection testing pass from the testing plan.

### DIFF
--- a/tinytorch/src/08_training/08_training.py
+++ b/tinytorch/src/08_training/08_training.py
@@ -1292,8 +1292,24 @@ def trainer_save_checkpoint(self, path: str):
     }
 
     Path(path).parent.mkdir(parents=True, exist_ok=True)
-    with open(path, 'wb') as f:
-        pickle.dump(checkpoint, f)
+    # Write to a temp file first, then atomically replace the target.
+    # Writing directly to `path` would leave a partially-written (corrupt)
+    # checkpoint in place if the process is killed, the disk fills up, or
+    # pickling fails partway through, silently destroying whatever good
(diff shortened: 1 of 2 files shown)


profvjreddi:

Two adjustments. The APPROACH now describes the temp-file-and-replace write, so a learner isn't told to use pickle.dump() directly and then failed for doing it. And I loosened your two temp-file assertions off the literal ".tmp" name onto the directory itself, empty after a failed save, only the checkpoint after a good one, so any correct atomic write passes rather than only this one.

PR #2070, opened 2026-08-17
fix(tinytorch): get_progress_data() crashed on wrong-type progress.json, silently swallowed malformed milestones.json

Shashank-Tripathi-07:
Part of #2046.

Same bug class already found and fixed twice in `tito/commands/milestone.py`: `json.load()` succeeds on syntactically valid JSON of the wrong shape (a bare list or string instead of the expected object), and code that assumes a dict crashes.

`get_progress_data()` is the single source of progress data for 14+ call sites across the module workflow (`is_module_started`, `mark_module_started`, `update_progress`, `get_last_worked_module`, ...), so an unvalidated wrong-type `progress.json` broke essentially the entire `tito module` command surface, not one isolated command. Confirmed the crash directly before fixing: `is_module_started()` raised a raw `AttributeError`.

`_check_milestone_unlocks()` had the same gap for `milestones.json`, though its own outer exception handler already prevented a crash from reaching the caller (by design, so milestone-check failures don't fail the whole workflow). That made it a silent-failure bug rather than a crash: a malformed `milestones.json` meant the function gave up early instead of treating it as empty and correctly recording newly-unlocked milestones. Fixed the same way, and added explicit UTF-8 encoding to both read and write sites in this file for consistency with the already-fixed `milestone.py`.

Verified via hand-mutation: reverting either isinstance check reproduces its respective failure mode (a hard crash for `progress.json`, a silently incomplete `milestones.json` for the other).

Found during a systematic audit of `tito/` for this bug pattern.

### DIFF
--- a/tinytorch/tests/cli/test_workflow_malformed_progress.py
+++ b/tinytorch/tests/cli/test_workflow_malformed_progress.py
@@ -1,0 +1,154 @@
+"""
+Malformed .tito/progress.json and .tito/milestones.json tests for
+ModuleWorkflowCommand (the `tito module start/resume/complete` workflow).
+
+Same bug class already found and fixed twice in tito/commands/milestone.py:
+json.load() succeeds on syntactically valid JSON of the wrong shape (a
+bare list instead of the expected object), and code that assumes a dict
+crashes with a raw AttributeError/TypeError deep inside whichever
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2071, opened 2026-08-17
fix(tinytorch): tito setup crashed completely on corrupted profile.json

Shashank-Tripathi-07:
Part of #2046.

`create_user_profile()` read `~/.tinytorch/profile.json` with zero exception handling at all, not even a narrow catch: a corrupted file (interrupted write, bad merge, manual edit gone wrong) crashed `tito setup` completely, with no way to recover short of manually deleting the file. Confirmed the crash directly before fixing (`JSONDecodeError`, uncaught).

The write side wasn't atomic either, the exact fault-injection pattern already fixed for `Trainer.save_checkpoint`: a crash mid-write to `profile.json` would corrupt it, which is the very thing the read side then couldn't recover from.

Fixed both: the read now catches `JSONDecodeError`/`IOError`/`UnicodeDecodeError` and validates the loaded value is actually a dict, falling back to creating a fresh profile rather than crashing setup entirely. The write now goes through a temp file and `os.replace()`, atomic on both POSIX and Windows, with cleanup on failure.

Verified via hand-mutation: reverting to the unguarded read reproduces both a `JSONDecodeError` crash (malformed JSON) and a `TypeError` crash (valid JSON, wrong shape).

Found during a systematic audit of `tito/` for this bug pattern.

### DIFF
--- a/tinytorch/tests/cli/test_setup_profile_malformed.py
+++ b/tinytorch/tests/cli/test_setup_profile_malformed.py
@@ -1,0 +1,87 @@
+"""
+Malformed/corrupted ~/.tinytorch/profile.json tests for
+SetupCommand.create_user_profile().
+
+create_user_profile() previously read profile.json with zero exception
+handling at all, not even a narrow catch: a corrupted file (bad merge,
+interrupted write, manual edit gone wrong) crashed `tito setup`
+completely, with no way to recover short of manually deleting the file.
+The write side wasn't atomic either, so an interrupted write could be
+the very thing that corrupts the file in the first place.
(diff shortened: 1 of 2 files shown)


profvjreddi:
Thanks again @Shashank-Tripathi-07. Integrated into dev.

This overlapped #2089 in setup.py, so a heads up on how I resolved it: I took your side at both hunks, since yours already carried the utf-8 encoding, and confirmed nothing from #2089 was lost.

Same as #2089 on the badge, it'll show as closed rather than merged.

PR #2072, opened 2026-08-17
fix(tinytorch): SubmissionHandler._read_json_safe returned wrong-type JSON as-is

Shashank-Tripathi-07:
Part of #2046.

Same bug class already found and fixed in `tito/commands/milestone.py` and `tito/commands/module/workflow.py`: `json.load()` succeeds on syntactically valid JSON of the wrong shape, and code that assumes a dict crashes. `_read_json_safe()`'s own docstring and type signature promise a `Dict[str, Any]` but never validated it, so a wrong-type `progress.json` or `milestones.json` (a bad merge, partial write) made `assemble_payload()` crash with a raw `AttributeError`, blocking `tito community sync`/submission entirely. Also added `UnicodeDecodeError` to the caught exceptions, since `encoding='utf-8'` was already explicit here but a genuinely non-UTF-8 file would still have escaped uncaught.

Confirmed the crash directly before fixing.

Verified via hand-mutation: reverting the isinstance check reproduces the crash for both `progress.json` and `milestones.json`.

Found during a systematic audit of `tito/` for this bug pattern (this is the fourth location it's been found in).

### DIFF
--- a/tinytorch/tests/cli/test_submission_malformed_json.py
+++ b/tinytorch/tests/cli/test_submission_malformed_json.py
@@ -1,0 +1,83 @@
+"""
+Malformed progress.json/milestones.json tests for
+SubmissionHandler._read_json_safe() / assemble_payload().
+
+Same bug class already found and fixed in tito/commands/milestone.py and
+tito/commands/module/workflow.py: json.load() succeeds on syntactically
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2073, opened 2026-08-17
fix(tinytorch): load_credentials() returned wrong-type JSON as-is, crashing login checks

Shashank-Tripathi-07:
Part of #2046.

Same bug class already found and fixed in `tito/commands/milestone.py`, `tito/commands/module/workflow.py`, and `tito/core/submission.py`: `json.load()` succeeds on syntactically valid JSON of the wrong shape (manual tampering, a bad merge), and code that assumes a dict crashes.

Every caller (`get_token`, `is_logged_in`, `get_user_email`, `get_refresh_token`) guards with `if creds:` before calling `.get()`, but a non-empty wrong-type value (a list, a non-empty string, a number) is truthy, so that guard doesn't actually protect against the crash. `is_logged_in()` is used throughout the CLI to decide whether to prompt for login, so this could break authentication entirely for anyone with a manually-edited or corrupted `credentials.json`.

Confirmed the crash directly before fixing (`AttributeError` on `get_token()`).

Verified via hand-mutation: reverting the isinstance check reproduces the crash across all four public helper functions.

Found during a systematic audit of `tito/` for this bug pattern (this is the fifth location it's been found in).

### DIFF
--- a/tinytorch/tests/cli/test_auth_malformed_credentials.py
+++ b/tinytorch/tests/cli/test_auth_malformed_credentials.py
@@ -1,0 +1,75 @@
+"""
+Malformed ~/.tinytorch/credentials.json tests for load_credentials()
+and its callers (get_token, is_logged_in, get_user_email,
+get_refresh_token).
+
+Same bug class already found and fixed in tito/commands/milestone.py,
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2074, opened 2026-08-17
fix(tinytorch): _offer_post_login_sync crashed on wrong-type progress.json

Shashank-Tripathi-07:
Part of #2046.

Same bug class already found and fixed in `tito/commands/milestone.py`, `tito/commands/module/workflow.py`, `tito/core/submission.py`, and `tito/core/auth.py`: `json.loads()` succeeds on syntactically valid JSON of the wrong shape (a bad merge, partial write), and `.get()` on the wrong type raises `AttributeError`, which wasn't one of the exceptions this call site caught (only `JSONDecodeError` and `OSError`). Also added explicit `encoding='utf-8'` to the read, matching every other fix in this series.

Confirmed the crash directly before fixing: a wrong-type `progress.json` crashed `tito login`'s post-login sync offer.

Verified via hand-mutation: reverting the isinstance check reproduces the crash.

Found during a systematic audit of `tito/` for this bug pattern (this is the sixth location it's been found in, all fixed in this initiative: #2049, #2070, #2072, #2073, and this one).

### DIFF
--- a/tinytorch/tests/cli/test_login_malformed_progress.py
+++ b/tinytorch/tests/cli/test_login_malformed_progress.py
@@ -1,0 +1,55 @@
+"""
+Malformed .tito/progress.json test for
+LoginCommand._offer_post_login_sync().
+
+Same bug class already found and fixed in tito/commands/milestone.py,
+tito/commands/module/workflow.py, tito/core/submission.py, and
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2075, opened 2026-08-17
fix(tinytorch): _prepare_notebook_for_nbgrader crashed on wrong-type notebook JSON

Shashank-Tripathi-07:
Part of #2046.

Same bug class already found and fixed across `tito/`: `json.loads()` succeeds on syntactically valid JSON of the wrong shape (a bad merge, interrupted write, manually-edited `.ipynb` gone wrong), and code that assumes a dict crashes. This function already validated that `notebook["cells"]` is a list, but not that `notebook` itself is a dict first, so `notebook.get("cells")` crashed with a raw `AttributeError` before that existing check ever ran.

Confirmed the crash directly before fixing.

Verified via hand-mutation: reverting the new dict-shape check reproduces the crash; a regression test confirms the pre-existing "missing cells list" validation still fires correctly for a valid dict that's just missing the `cells` key.

Found during a systematic audit of `tito/` for this bug pattern (the seventh and final location found; the remaining `json.load` call sites in `tito/` were checked and confirmed already safe, either broadly caught or parsing trusted subprocess/API output rather than user-editable state).

### DIFF
--- a/tinytorch/tests/cli/test_nbgrader_notebook_malformed.py
+++ b/tinytorch/tests/cli/test_nbgrader_notebook_malformed.py
@@ -1,0 +1,68 @@
+"""
+Malformed notebook JSON tests for
+NBGraderCommand._prepare_notebook_for_nbgrader().
+
+Same bug class already found and fixed across tito/: json.loads()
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2076, opened 2026-08-17
test(tinytorch): extend PyTorch cross-checking to Linear layer and losses

Shashank-Tripathi-07:
Part of #2046.

Extends the existing tensor/autograd cross-check suite (#2047, which added the optional `cross-check` dependency group) to `Linear`, `MSELoss`, `CrossEntropyLoss`, and `BinaryCrossEntropyLoss`, none of which had ever been directly compared against PyTorch's own implementations. Finite-difference and mutation testing can only prove internal consistency; only comparison against an independent reference can catch a bug that's consistently wrong in both forward and backward.

All 25 test executions (5 random seeds x 5 test cases) pass: `Linear`'s forward and backward match `torch.nn.Linear` (accounting for the transposed weight-storage convention between the two), and all three losses match their `torch.nn.functional` equivalents exactly.

### DIFF
--- a/tinytorch/pyproject.toml
+++ b/tinytorch/pyproject.toml
@@ -45,6 +45,14 @@ dev = [
     "ipykernel>=7.2.0",
     "nbdev>=3.0.15,<3.0.16",
 ]
+cross-check = [
+    # Test-only dependency for tests/cross_check/, never imported by the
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2077, opened 2026-08-17
test(tinytorch): extend PyTorch cross-checking to Conv2d and attention

Shashank-Tripathi-07:
Part of #2046.

Extends the cross-check suite to `Conv2d` and `scaled_dot_product_attention`, neither of which had ever been directly compared against PyTorch's own implementations.

All 20 test executions (5 random seeds x 4 test cases) pass: `Conv2d`'s forward and backward match `torch.nn.Conv2d` exactly (same weight-storage convention, no transpose needed), and `scaled_dot_product_attention`'s forward output/weights and backward gradients match a from-scratch PyTorch reference implementation of the same formula.

### DIFF
--- a/tinytorch/pyproject.toml
+++ b/tinytorch/pyproject.toml
@@ -45,6 +45,14 @@ dev = [
     "ipykernel>=7.2.0",
     "nbdev>=3.0.15,<3.0.16",
 ]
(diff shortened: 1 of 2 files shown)


Shashank-Tripathi-07:

PR #2078, opened 2026-08-17
test(tinytorch): close three exact-value gaps in activations found via mutation testing

Shashank-Tripathi-07:
Part of #2046.

Found via mutation testing (cosmic-ray) against `tinytorch/core/activations.py`:

1. **GELU had zero pytest coverage at all** (only a loose-bound inline test that pytest never collects, checking e.g. `result > 0.8` rather than the exact value), so its formula (`x * sigmoid(1.702 * x)`) was never verified precisely: mutating the 1.702 constant, or either multiplication to addition, all survived mutation testing undetected.

2. **Softmax's default `dim=-1` parameter was never exercised** on a multi-dimensional, non-symmetric tensor. Also discovered while fixing this: `Softmax.forward()`'s own default is never actually reachable once autograd is enabled (`enable_autograd()` replaces it entirely with `tracked_softmax_forward`, defined separately in `06_autograd.py` with its own independently hardcoded `dim=-1` default), the same monkeypatch-shadowing architecture already found and fixed elsewhere in this codebase (#2051). Both defaults currently agree so there's no live bug, but documented the constraint in the test so it isn't rediscovered from scratch.

3. **Sigmoid's numerically-stable branch selection** (`x_data >= 0`) was never tested directly through `activations.Sigmoid` at a magnitude where the unstable branch would actually overflow; only the autograd-tracked path had this coverage.

Verified via hand-mutation on all three.

### DIFF
--- a/tinytorch/tests/02_activations/test_activations_exact_value_gaps.py
+++ b/tinytorch/tests/02_activations/test_activations_exact_value_gaps.py
@@ -1,0 +1,109 @@
+"""
+Exact-value and boundary tests closing three gaps found via mutation
+testing in tinytorch.core.activations:
+
+1. GELU had zero pytest coverage at all (only a loose-bound inline test
+   that pytest never collects), so its exact formula
+   (x * sigmoid(1.702 * x)) was never verified precisely: mutating the
+   1.702 constant, or either multiplication to addition, all survived.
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2079, opened 2026-08-17
test(tinytorch): add exact-behavior coverage for Dropout

Shashank-Tripathi-07:
Part of #2046.

Mutation testing against `layers.py` found `Dropout` essentially uncovered: the only existing test elsewhere in the suite accepts "output changed OR equals x*2" as sufficient, not strong enough to catch a wrong scale factor or a wrong drop-probability comparison.

Covers `_should_apply_dropout`'s boundary conditions (`training=False`, `p=0`), `_generate_dropout_mask`'s exact scaling (kept elements must equal exactly `1/(1-p)`, not any other factor) and statistical drop rate, and `forward()`'s `p=1.0` special case and eval-mode passthrough, none previously exercised precisely.

Verified via hand-mutation: an AND-to-OR mutation in `_should_apply_dropout` and a wrong scale formula are both caught; a strict-to-non-strict comparison mutation in the random threshold (`<` vs `<=`) is confirmed to be a true equivalent mutant for a continuous uniform draw (P(exact equality) = 0), not a real gap.

### DIFF
--- a/tinytorch/tests/03_layers/test_dropout_exact_behavior.py
+++ b/tinytorch/tests/03_layers/test_dropout_exact_behavior.py
@@ -1,0 +1,121 @@
+"""
+Exact-behavior tests for Dropout, found via mutation testing to have
+essentially no coverage (only one loose existing test, elsewhere in the
+suite, that accepts "output changed OR equals x*2" as sufficient, not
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2082, opened 2026-08-18
Add direct test coverage for log_softmax (previously untested)

Shashank-Tripathi-07:
## What

`log_softmax` (used internally by `CrossEntropyLoss`) had zero direct test coverage anywhere in the suite. The only test with "log softmax" in its name, `test_unit_log_softmax` in `tests/integration/test_optimizers_integration.py`, never actually calls `log_softmax`, it exercises the `Softmax` activation class instead:

```python
def test_unit_log_softmax():
    """Test log softmax computation."""
    x = Tensor(np.array([[1.0, 2.0, 3.0]]))
    softmax = Softmax()
    output = softmax(x)
    assert np.isclose(output.data.sum(), 1.0, atol=1e-5)
```

## Why

Found while triaging mutation-testing survivors for `04_losses`. Four survivor mutations inside `log_softmax` were unkillable: both `keepdims=True` flags flipped to `False`, a sign flip in the final `log_sum_exp` subtraction, and the `dim=-1` default's unary minus, none of which affected any test outcome anywhere in the suite.

## Fix

Added `tests/04_losses/test_log_softmax_direct.py` with direct tests for:
- exact value against a manual `log(softmax(x))` reference (1D and 2D)
- output validity (exponentiated result sums to 1 per row)
- `keepdims` correctness on a non-square 2D input across both `dim=-1` and `dim=0`
- numerical stability at overflow/underflow-triggering magnitudes
- the default `dim=-1` parameter

- `keepdims=True` -> `False` on the max reduction: **caught** (5 test failures)
- `log_sum_exp` sign flip (`- max_vals - log_sum_exp` -> `- max_vals + log_sum_exp`): **caught** (6 test failures)
- `dim=-1` default -> `dim=1` default: **not caught, and correctly so** — this is a true equivalent mutant for this codebase. `axis=-1` and `axis=1` are identical for any 2D tensor, and `log_softmax` is only ever called on 2D `(batch, classes)` logits here, so the default's specific value can't be distinguished by any realistic test. Documented in the test module docstring rather than chased further.

All 9 new tests pass on unmutated code; full `tests/04_losses` suite (24 tests) still passes.

### DIFF
--- a/tinytorch/tests/04_losses/test_log_softmax_direct.py
+++ b/tinytorch/tests/04_losses/test_log_softmax_direct.py
@@ -1,0 +1,123 @@
+"""
+Module 04: Losses - Direct log_softmax Tests
+=============================================
+
+log_softmax is a standalone helper used internally by CrossEntropyLoss,
+but had zero direct test coverage: the only test named "log softmax" in
+the suite (tests/integration/test_optimizers_integration.py::test_unit_log_softmax)
+actually exercises the Softmax activation class, never calls log_softmax at all.
+
+These tests exercise log_softmax directly, covering:
+- exact numerical output against a manual log(softmax(x)) reference
+- the default dim=-1 parameter
+- keepdims correctness on a multi-row 2D input (dim=-1 and dim=0)
+- numerical stability at overflow-triggering magnitudes
+"""
+
+import numpy as np
+import pytest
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2083, opened 2026-08-18
Add Tensor-gradient test coverage for clip_grad_norm

Shashank-Tripathi-07:
## What

`clip_grad_norm`'s own docstring example assigns `.grad` as a `Tensor`:

```python
>>> params[0].grad = Tensor([10, 20, 30])  # Large gradients
>>> original_norm = clip_grad_norm(params, max_norm=1.0)
```

and the implementation has a dedicated `isinstance(param.grad, np.ndarray)` branch specifically to support both `Tensor` and plain-`ndarray` gradients:

```python
if isinstance(param.grad, np.ndarray):
    param.grad = param.grad * clip_coef
else:
    # Trust that Tensor has .data attribute
    param.grad.data = param.grad.data * clip_coef
```

Every existing test in `TestClipGradNorm` (`tests/08_training/test_training_coverage.py`) only ever constructs params with a plain `ndarray` `.grad`, so the documented `Tensor`-gradient branch was completely untested.

## Why

Found while triaging mutation-testing survivors for `08_training`. Line 401 (`param.grad.data = param.grad.data * clip_coef`, the `Tensor` branch's scaling step) had 11 unkillable survivor mutants, all because that branch was never exercised by any test.

## Fix

Added `TestClipGradNormTensorGradients`, mirroring the existing `ndarray`-gradient test cases (`test_returns_original_norm`, `test_clips_large_gradients`, `test_does_not_clip_small_gradients`, `test_clips_across_multiple_params`) but with `Tensor`-typed `.grad`, plus `test_mixed_ndarray_and_tensor_gradients_share_one_global_norm` confirming the global norm is computed across both representations together and each is scaled correctly in its own type.

`param.grad.data = param.grad.data * clip_coef` mutated to `+ clip_coef`: previously had zero effect on any test; now caught by 3 of the new tests.

### DIFF
--- a/tinytorch/tests/08_training/test_training_coverage.py
+++ b/tinytorch/tests/08_training/test_training_coverage.py
@@ -153,6 +153,63 @@ class TestClipGradNorm:
         np.testing.assert_allclose(params[0].grad, np.zeros(3))
 
 
+class TestClipGradNormTensorGradients:
+    """clip_grad_norm's own docstring example assigns .grad as a Tensor
+    (`params[0].grad = Tensor([10, 20, 30])`), and the implementation has a
+    dedicated isinstance(param.grad, np.ndarray) branch specifically to
+    support both Tensor and plain-ndarray gradients. Every test above only
+    ever exercises the ndarray branch, leaving the Tensor branch completely
+    untested despite being the documented usage pattern."""
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2084, opened 2026-08-18
Add test coverage for train_epoch's gradient accumulation remainder

Shashank-Tripathi-07:
## What

`Trainer.train_epoch`'s in-loop update fires when `(batch_idx + 1) % accumulation_steps == 0`. When the number of batches is not a multiple of `accumulation_steps`, a final partial group is left over and must be flushed by the post-loop `if accumulated_loss > 0` branch:

```python
for batch_idx, (inputs, targets) in enumerate(dataloader):
    accumulated_loss += self._process_batch(inputs, targets, accumulation_steps)
    if (batch_idx + 1) % accumulation_steps == 0:
        self._optimizer_update()
        total_loss += accumulated_loss
        accumulated_loss = 0.0
        num_batches += 1
        self.step += 1

# Handle remaining accumulated gradients
if accumulated_loss > 0:
    self._optimizer_update()
    total_loss += accumulated_loss
    num_batches += 1
```

Every existing accumulation test in this suite (`tests/08_training/test_training_core.py`, `tests/08_training/test_training_coverage.py`) used batch counts that divide evenly by `accumulation_steps`, so the remainder branch, and the `total_loss`/`num_batches` accounting around it, was never exercised.

## Why

Found while triaging mutation-testing survivors for `08_training`. The remainder condition (line 381 in the exported package) had 8 unkillable survivor mutants, and the surrounding accounting (lines 373-389) had dozens more.

## Fix

Added `TestGradientAccumulationRemainder`:
- a leftover batch still triggers a weight update
- `avg_loss` divides by the update-group count, not the raw batch count
- an epoch made entirely of one partial group (no full group ever forms)
- `train_loss` history bookkeeping stays correct across a remainder epoch
- the boundary case in the other direction: when the epoch ends exactly on a full accumulation group, the remainder branch must be a true no-op, not fire an extra spurious update

`if accumulated_loss > 0` mutated to `>= 0`: previously had zero effect on any test. Now caught, the spurious-update boundary test fails because the mutation fires a phantom update group that incorrectly halves `avg_loss`.

### DIFF
--- a/tinytorch/tests/08_training/test_training_coverage.py
+++ b/tinytorch/tests/08_training/test_training_coverage.py
@@ -445,6 +445,94 @@ class TestGradientClippingIntegration:
 
 
 # ─────────────────────────────────────────────
+# Gradient accumulation remainder handling
+# ─────────────────────────────────────────────
+
+class TestGradientAccumulationRemainder:
+    """train_epoch's in-loop update fires when (batch_idx + 1) % accumulation_steps == 0.
+    When the number of batches is not a multiple of accumulation_steps, a final
+    partial group is left over and must be flushed by the post-loop
+    'if accumulated_loss > 0' branch. Every existing accumulation test in this
+    suite uses batch counts that divide evenly, so that remainder branch, and
+    the total_loss/num_batches accounting around it, was never exercised."""
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2085, opened 2026-08-18
Add one-hot target test coverage for Trainer.evaluate accuracy

Shashank-Tripathi-07:
## What

`Trainer.evaluate`'s accuracy computation branches on whether targets are integer class indices (1D) or one-hot encoded (2D), with its own comment documenting both as supported:

```python
if len(targets.data.shape) == 1:  # Integer targets
    correct += np.sum(predictions == targets.data)
else:  # One-hot targets
    correct += np.sum(predictions == np.argmax(targets.data, axis=1))
```

Every existing classification test in this suite (`tests/08_training/test_training_coverage.py`) used integer targets only, leaving the one-hot branch completely untested.

## Why

Found while triaging mutation-testing survivors for `08_training`. That line had 9 unkillable survivor mutants.

## A note on scope

`CrossEntropyLoss.forward` only accepts integer class-index targets per its own docstring (`targets.data.astype(int)` on a one-hot 2D array produces nonsense), so it can't actually be paired with one-hot targets in practice. The new tests use a minimal stub loss to isolate `evaluate()`'s own accuracy logic from that unrelated limitation, since `evaluate()`'s one-hot handling exists specifically for the accuracy computation and is independent of which loss function is in use.

## Fix

Added:
- `test_classification_accuracy_with_onehot_targets` / `test_zero_accuracy_for_wrong_predictions_onehot_targets`: one-hot counterparts of the existing integer-target perfect/wrong-classifier tests
- `test_mixed_onehot_batch_partial_accuracy`: a batch with some correct and some incorrect one-hot predictions, exercising the `np.sum(...)` count itself (fractional accuracy) rather than only the two all-correct/all-wrong extremes

The one-hot argmax's `axis=1` flipped to `axis=0`: previously had zero effect on any test. Now caught (2 test failures, both surfacing the resulting shape-mismatch `ValueError`).

Full `tests/08_training` suite (62 tests) still passes.

### DIFF
--- a/tinytorch/tests/08_training/test_training_coverage.py
+++ b/tinytorch/tests/08_training/test_training_coverage.py
@@ -352,6 +352,99 @@ class TestTrainerEvaluate:
         _, accuracy = trainer.evaluate(data)
         assert accuracy == 0.0, f"Wrong classifier should have accuracy=0.0, got {accuracy}"
 
+    def test_classification_accuracy_with_onehot_targets(self):
+        """evaluate() must handle one-hot encoded targets (2D, shape (N, C)),
+        not just integer class-index targets (1D). Every other classification
+        test in this suite uses integer targets, leaving evaluate's one-hot
+        branch (targets.data.shape has ndim > 1) completely untested.
+
+        Uses a stub loss (CrossEntropyLoss itself only accepts integer class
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2086, opened 2026-08-18
Add scheduler checkpoint round-trip test coverage

Shashank-Tripathi-07:
## What

`_get_scheduler_state` / `_set_scheduler_state` serialize and restore the scheduler's `max_lr`/`min_lr`/`total_epochs` through `save_checkpoint` / `load_checkpoint`, including `None`-guard branches for the scheduler-absent case on both the save and load sides:

```python
def _set_scheduler_state(self, state):
    if state is None or self.scheduler is None:
        return
    for key, value in state.items():
        if hasattr(self.scheduler, key):
            setattr(self.scheduler, key, value)
```

No existing checkpoint test (`tests/08_training/test_training_coverage.py::TestCheckpointing`) used a `Trainer` with a scheduler attached at all, leaving this entire round-trip untested.

## Why

Found while triaging mutation-testing survivors for `08_training`. `_set_scheduler_state`'s guard had 4 unkillable survivor mutants, plus more across `_get_scheduler_state` and the attribute restore loop.

## Fix

Added `TestSchedulerCheckpointState`:
- attribute round-trip (`max_lr`/`min_lr`/`total_epochs` survive save → corrupt → load)
- the restored scheduler produces the same `get_lr()` values across a full epoch range, not just matching attributes by coincidence
- `scheduler_state` is correctly saved as `None` when no scheduler is attached
- both cross-directions of a scheduler/no-scheduler mismatch between save time and load time: a checkpoint saved *without* a scheduler loaded into a trainer that *has* one (must leave that scheduler untouched), and the reverse (must not crash)

The guard's `or` flipped to `and`: previously had zero effect on any test. Now caught, the state-is-`None`-but-scheduler-present case correctly raises an `AttributeError` before the fix's `None`-check could short-circuit it.

### DIFF
--- a/tinytorch/tests/08_training/test_training_coverage.py
+++ b/tinytorch/tests/08_training/test_training_coverage.py
@@ -260,6 +260,105 @@ class TestCheckpointing:
             assert os.path.exists(path)
 
 
+class TestSchedulerCheckpointState:
+    """_get_scheduler_state / _set_scheduler_state round-trip through
+    save_checkpoint / load_checkpoint. No existing checkpoint test used a
+    Trainer with a scheduler attached, leaving this entire round-trip, and
+    the None-guard branches on both sides, completely untested."""
+
+    def test_scheduler_attributes_round_trip(self):
+        scheduler = CosineSchedule(max_lr=0.1, min_lr=0.01, total_epochs=20)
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2087, opened 2026-08-18
Add momentum buffer checkpoint round-trip test coverage

Shashank-Tripathi-07:
## What

`_get_optimizer_state` / `_set_optimizer_state` conditionally include momentum buffers in the checkpoint dict, gated on `optimizer.has_momentum()`:

```python
def _get_optimizer_state(self):
    state = {}
    state['lr'] = self.optimizer.lr
    if hasattr(self.optimizer, 'has_momentum') and self.optimizer.has_momentum():
        momentum_state = self.optimizer.get_momentum_state()
        if momentum_state is not None:
            state['momentum_buffers'] = momentum_state
    return state

def _set_optimizer_state(self, state):
    if 'lr' in state:
        self.optimizer.lr = state['lr']
    if 'momentum_buffers' in state:
        if hasattr(self.optimizer, 'has_momentum') and self.optimizer.has_momentum():
            self.optimizer.set_momentum_state(state['momentum_buffers'])
```

Every existing checkpoint test (`tests/08_training/test_training_coverage.py::TestCheckpointing`) used a plain `SGD` with `momentum=0.0` (the default), so the `has_momentum()==True` branch was never exercised on either the save or load side, and momentum buffers were never actually round-tripped.

## Why

Found while triaging mutation-testing survivors for `08_training`. This cluster had 8 unkillable survivor mutants combined across `_get_optimizer_state` and `_set_optimizer_state`.

## Fix

Added `TestOptimizerMomentumCheckpointState`:
- `momentum_buffers` key present in the saved state when momentum is enabled
- key absent when it isn't (SGD with `momentum=0.0`)
- actual buffer values (not just key presence) survive a corrupt-then-restore cycle
- a training-behavior-level check: a trainer restored with prior momentum takes a genuinely different next optimizer step than one starting momentum from zero, isolating the checkpoint logic from a coincidentally-passing "looks equal" comparison

The momentum buffer restore guard flipped from `'momentum_buffers' in state` to `'momentum_buffers' not in state`: previously had zero effect on any test. Now caught (2 test failures).

Full `tests/08_training` suite (63 tests) still passes.

### DIFF
--- a/tinytorch/tests/08_training/test_training_coverage.py
+++ b/tinytorch/tests/08_training/test_training_coverage.py
@@ -260,6 +260,108 @@ class TestCheckpointing:
             assert os.path.exists(path)
 
 
+class TestOptimizerMomentumCheckpointState:
+    """_get_optimizer_state / _set_optimizer_state conditionally include
+    momentum buffers, gated on optimizer.has_momentum(). Every existing
+    checkpoint test uses a plain SGD with momentum=0 (the default), so the
+    has_momentum()==True branch on both the save and load sides was never
+    exercised, and the momentum buffers were never actually round-tripped."""
+
+    def _trainer_with_momentum(self, momentum=0.9, lr=0.1):
+        model = simple_model()
+        opt = SGD(model.parameters(), lr=lr, momentum=momentum)
+        return Trainer(model, opt, MSELoss()), model, opt
+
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:

PR #2088, opened 2026-08-18
Fix Windows cp1252 crashes in subprocess-based tests

Shashank-Tripathi-07:
## What

`subprocess.run(..., text=True)` without an explicit `encoding=` decodes the child process's stdout/stderr using `locale.getpreferredencoding()`, which is cp1252 on Windows. `tito`'s own CLI output (via `rich`) contains characters that aren't valid cp1252, so the internal stdout reader thread crashed with an unhandled `UnicodeDecodeError`:

```
File "...\Lib\subprocess.py", line 1614, in _readerthread
  buffer.append(fh.read())
File "...\Lib\encodings\cp1252.py", line 23, in decode
  return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 607: character maps to <undefined>
```

leaving `result.stdout` as `None`, which then crashed one line later in whatever test tried to search it (`TypeError: argument of type 'NoneType' is not a container or iterable`). Several `Path.read_text()` / `open()` calls in the same test files had the identical default-encoding gap.

## Why this matters

This was previously worked around by excluding `tests/cli`, `tests/e2e`, `tests/environment`, and `tests/milestones/test_milestones_run.py` from "full suite" runs on Windows, treating the failures as environmental noise. They aren't: they're entirely caused by a missing `encoding=` argument, not by anything wrong with the commands or files under test, and are fixable with no behavior change on Linux/Mac (where the default encoding is already UTF-8, so this is a no-op there).

## Fix

Added `encoding='utf-8', errors='replace'` to all 42 `subprocess.run(..., text=True)` call sites across `tests/cli`, `tests/environment`, `tests/e2e`, `tests/integration`, and `tests/milestones`, and `encoding='utf-8'` to the handful of `read_text()`/`open()` calls in the same files. Also extended `tests/milestones/milestone_tracker.py`'s `except` clauses to catch `UnicodeDecodeError`, matching the same fix pattern already applied across `tito/` earlier this session (#2049, #2070-2075).

## Testing

- `tests/cli`: 92 passed, 1 skipped (was 13 failed, 79 passed, 1 skipped)
- `tests/environment`: 58 passed, 1 failed (was multiple cp1252-related failures). The one remaining failure (`test_pip_available`) is unrelated: this dev venv genuinely has no `pip` module installed (created via `uv`), not an encoding issue, so left as-is.

### DIFF
--- a/tinytorch/tests/cli/test_cli_execution.py
+++ b/tinytorch/tests/cli/test_cli_execution.py
@@ -32,7 +32,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         # Should exit successfully
@@ -48,7 +48,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', '-h'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         assert result.returncode == 0
@@ -62,7 +62,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', '--version'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         assert result.returncode == 0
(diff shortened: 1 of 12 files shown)


Shashank-Tripathi-07:
Superseded by #2089, which includes this PR's test-file changes plus the same encoding fix applied to tito/'s own source files. Closing in favor of that one to avoid overlapping review.

PR #2089, opened 2026-08-18
Fix Windows cp1252 encoding crashes across all tito file I/O (tests + source)

Shashank-Tripathi-07:
## What

Supersedes #2088 (closing it in favor of this PR, which is a superset). Fixes the Windows cp1252-default-encoding problem everywhere it appears in `tito/`, not just in tests: every `subprocess.run(..., text=True)` call, and every plain `open()` / `read_text()` / `write_text()` call, that was missing an explicit `encoding=`.

## Root cause

On Windows, `subprocess.run(..., text=True)` without an explicit `encoding=` decodes the child process's stdout/stderr using `locale.getpreferredencoding()`, which is cp1252. `open()`/`read_text()`/`write_text()` without `encoding=` behave the same way. `tito`'s own CLI output (via `rich`) and various progress/config JSON files can contain characters outside cp1252, so reads/writes/subprocess captures crash with an unhandled `UnicodeDecodeError` or `UnicodeEncodeError`. On Linux/Mac this is a no-op fix, since the default encoding there is already UTF-8.

## Scope

**Tests** (`tests/cli`, `tests/environment`, `tests/e2e`, `tests/integration`, `tests/milestones`): all 42 `subprocess.run(..., text=True)` call sites, plus a handful of `read_text()`/`open()` calls in the same files. Also extended `tests/milestones/milestone_tracker.py`'s `except` clauses to catch `UnicodeDecodeError`.

**`tito/` source** (the CLI itself, not just its tests): `subprocess.run`/`Popen` calls here were already correctly encoded. The gap was in plain file I/O:
- `benchmark.py`: results/config/profile file reads and writes
- `milestone.py`: progress and config file reads/writes (9 sites)
- `commands/module/reset.py`, `commands/package/reset.py`: progress/milestone reset writes
- `commands/module/workflow.py`: progress and milestones file reads/writes
- `commands/setup.py`: user profile read/write
- `commands/dev/preflight.py`: preflight log file write
- `commands/system/health.py`: Jupyter kernelspec JSON read
- `main.py`, `__init__.py`: `pyproject.toml` version parsing
- `commands/login.py`: post-sync progress check (also extended its `except` to catch `UnicodeDecodeError`)
- `core/virtual_env_manager.py`: venv config lookup
- `core/status_analyzer.py`: the dynamically generated import-check script, both the temp file write and the `exec(open(...).read())` call that runs inside the spawned subprocess
- `core/auth.py`, `core/browser.py`: `/proc/version` reads (Linux/WSL-only, fixed for consistency)

## What's intentionally NOT in scope

A handful of already-open PRs from earlier this session (#2073 auth.py, #2074 login.py, #2070 workflow.py, etc.) fix a *different*, related bug: `json.load()` succeeding on syntactically-valid-but-wrong-type JSON and crashing downstream with `AttributeError`. This PR only adds `encoding='utf-8'` and (where an existing broad `except` was already present) extends it to catch `UnicodeDecodeError`; it does not add `isinstance` validation or otherwise change control flow. There will be a small amount of surface overlap with those PRs on `login.py` (both touch the same read line for different reasons) that the maintainer can trivially resolve on merge.

## Testing

- `tests/cli`: 92 passed, 1 skipped (was 13 failed, 79 passed)
- `tests/environment`: 58 passed, 1 failed. That one remaining failure (`test_pip_available`) is unrelated: this dev venv has no `pip` module installed, nothing to do with encoding.
- Full suite excluding `tests/cli`/`tests/e2e`/`tests/environment`/`tests/milestones/test_milestones_run.py` (which still have a separate, unrelated `WinError 193` subprocess-executable-resolution issue not addressed here): 762 passed, 35 skipped, 0 new failures. The only failure seen was the same pre-existing order-dependent flake (`test_deep_network_gradient_chain`) already known from earlier in this testing effort, confirmed unrelated by rerunning `tests/integration` alone (all passing).
- All modified files verified with `py_compile`.

### DIFF
--- a/tinytorch/tests/cli/test_cli_execution.py
+++ b/tinytorch/tests/cli/test_cli_execution.py
@@ -32,7 +32,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         # Should exit successfully
@@ -48,7 +48,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', '-h'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         assert result.returncode == 0
@@ -62,7 +62,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', '--version'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         assert result.returncode == 0
@@ -78,7 +78,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', command, '-h'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         # Help should always succeed
@@ -106,7 +106,7 @@ class TestCommandExecution:
             [sys.executable, '-m', 'tito.main', command, subcommand, '-h'],
             cwd=self.project_root,
             capture_output=True,
-            text=True
+            text=True, encoding='utf-8', errors='replace'
         )
 
         # Subcommand help should work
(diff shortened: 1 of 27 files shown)


profvjreddi:

I checked the remaining unencoded read_text() calls in tito and they're the four you fixed, so this closes it out rather than leaving a tail.

On the badge: I applied your commit onto a review branch off current dev rather than merging the PR head, so GitHub will show this as closed instead of merged. Authorship and credit are intact.

PR #2090, opened 2026-08-18
changed location of profiling function in tinytorch module 14

vedant-a-joshi:
This PR is for TinyTorch Module 14.

This makes it such that `profile_forward_pass` goes immediately before `profile_backward_pass`, which follows the testing order and logical flow.

Checked manually using `tito module view 14` and `pytest 14_profiling/`.

### DIFF
--- a/tinytorch/src/14_profiling/14_profiling.py
+++ b/tinytorch/src/14_profiling/14_profiling.py
@@ -754,43 +754,6 @@ class Profiler:
         }
         ### END SOLUTION
 
-    def profile_forward_pass(self, model, input_tensor) -> Dict[str, Any]:
-        """
-        Comprehensive profiling of a model's forward pass.
-
-        TODO: Gather measurements, then use _compute_derived_metrics and _analyze_bottleneck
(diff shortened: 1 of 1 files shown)


profvjreddi:
Thanks @vedant-a-joshi. I landed this and then backed it out, so I owe you the reasoning.

The Profiler class is ordered as two matched beats: the helpers for a measurement, then the entry point that composes them. profile_forward_pass calls _compute_derived_metrics and _analyze_bottleneck and nothing else, and both were already directly above it. Moving it below _estimate_backward_costs and _estimate_optimizer_memory puts two helpers it never calls between it and the two it does, and collapses the forward/backward pairing into one flat block of helpers followed by one flat block of entry points.

Nothing here is a functional change, the whole region is a single notebook cell, so this is purely about what a learner reads top to bottom, and in a teaching module that ordering is doing real work.

Your #2093 was a genuinely good catch and it's in dev.

vedant-a-joshi:
Ok, thanks.

PR #2091, opened 2026-08-19
Fix dead-code and numerical-instability bugs across autograd's tracked forward implementations

Shashank-Tripathi-07:
## What

Combines 5 previously-separate PRs (#2048, #2051, #2081, #2040, #2060), all of which touch `tinytorch/src/06_autograd/06_autograd.py`'s `enable_autograd()` monkeypatch. Combining them avoids five separate PRs churning the same function.

## The shared bug class

`enable_autograd()` replaces several `Tensor`/`Sigmoid`/loss-class methods with "tracked" versions that add gradient bookkeeping. Some of these tracked versions **reimplemented** the original operation's math from scratch instead of delegating to the module's own already-correct implementation. This meant:

- The original implementation (what a student actually reads and is graded on) became **dead code** the moment autograd loads, which is always, since `enable_autograd()` runs unconditionally on package import.
- Any fix or improvement made to the original silently never shipped, since the live code path was the separate reimplementation.
- One of these reimplementations (`tracked_sigmoid_forward`) had actually drifted: it used the naive `1/(1+exp(-x))` formula instead of the original's numerically-stable, sign-branching version, causing spurious overflow warnings (and eventually NaN) at large-magnitude inputs common right after gradients explode during training.

## Changes (one per original commit)

1. **#2048** — `tracked_sigmoid_forward` used an unstable formula; fixed to delegate to the original numerically-stable `Sigmoid.forward`.
2. **#2051** (2 commits) — `__radd__`/`__rsub__`/`__rmul__`/`__rtruediv__`/`sum()`, and the scalar branch of `__add__`/`__sub__`/`__truediv__`, all reimplemented Module 01's arithmetic instead of delegating to it.
3. **#2081** — `MSELoss`/`CrossEntropyLoss`/`BinaryCrossEntropyLoss.forward` (in `04_losses.py`) were fully shadowed by reimplementations (`tracked_mse_forward` etc.) instead of delegating.
4. **#2040** — `CrossEntropyLoss.forward` used target indices in numpy fancy indexing with no bounds check, surfacing a raw `IndexError` instead of a clear domain error. Originally needed fixing in both the original and the autograd-tracked duplicate; now that #2081 (in this same combined PR) makes the tracked version delegate, it only needs fixing once.
5. **#2060** — `GELUBackward` overflowed on large-magnitude inputs, the same numerical-instability failure mode as #2048's sigmoid bug, in the same file.

## Testing

`tests/06_autograd`, `tests/04_losses`, `tests/02_activations`: 119 passed, 18 skipped. Full suite (excluding the known Windows-only-broken `tests/cli`/`tests/e2e`/`tests/environment`/`tests/milestones/test_milestones_run.py`): 768 passed, 35 skipped, 0 new failures (one pre-existing order-dependent flake, confirmed unrelated by rerunning `tests/integration` alone).


profvjreddi:
Thanks @Shashank-Tripathi-07. Integrated into dev, and this one paid for itself.

It fixed a test that was already red on dev before your PR: test_deep_network_gradient_chain in tests/integration/test_training_flow.py. The tracked_* wrappers shadowing the modules' real forward passes is exactly what that test was catching, so delegating to the captured originals closed it. That suite went from 774 passed / 1 failed to 790 passed / 0 failed.

I also added the target-index validation to the CrossEntropyLoss APPROACH and HINTS, same reason as #2036.

PR #2092, opened 2026-08-19
Fix Jupyter process leak and start/resume desync dead-end in module workflow

Shashank-Tripathi-07:
## What

Combines 2 previously-separate PRs (#2011, #2026), both of which touch `tito/commands/module/workflow.py`. Combining them avoids two PRs churning the same file back to back.

## Changes

1. **#2011** (4 commits, including 3 follow-up fixes to the liveness check itself) — Every `tito module start` (without `--no-jupyter`), `resume`, or `view` launched a brand new `jupyter lab` subprocess with no tracking, no cleanup, and no check for whether one was already running. Confirmed 45 leaked processes after routine module-by-module testing. Fixed by tracking the launched process's PID in `.tito/jupyter.pid` and reusing an already-running server instead of spawning a duplicate. The liveness check itself went through a few iterations during review/testing: `os.kill(pid, 0)` isn't reliable on Windows, so it moved to a stdlib-only cross-platform check, plus a follow-up fix for how the kernel-level PID value was being handled.

2. **#2026** — `tito module start` and `tito module resume` could deadlock each other when `.tito/progress.json`'s tracking (`started_modules`) desynced from the actual notebook on disk under `modules/`. Each command's failure message pointed at the other, and neither ever mentioned the actual fix. `start_module` and `resume_module` now check whether the notebook directory actually exists before trusting the tracking flag, and recreate it from `src/` when it doesn't, instead of dead-ending.

## Testing

Both fixes were manually verified against a fresh `dev` install (process-count checks via process listing for #2011, notebook-recreation checks for #2026) at the time each was originally written. Recombined here with no logic changes; `py_compile` clean, no merge conflicts across the two originals (fully sequential, same-file changes).


profvjreddi:
Thanks @Shashank-Tripathi-07. Integrated into dev, CI green.

The Windows note in your PR was right, and it turns out the POSIX side needed the same care. `/proc` is Linux-only, so on macOS the fallback returned True for anything still alive:

```
_pid_is_running_jupyter(os.getpid())   # the running python interpreter
True
```

Since the pid file outlives the server it names and pids get recycled, that meant `tito module start` could report "Jupyter Lab is already running" against an unrelated process and never launch one again.

Fixed by falling back to `ps -p <pid> -o command=`, which macOS and the BSDs provide, and returning False when the probe fails: an extra Jupyter process is a cheaper wrong answer than a student who can never start one. Also deleting the pid file once its process is gone, and pointing the reuse message at localhost:8888 the way the launch path does.

PR #2093, opened 2026-08-21
tinytorch module 14: add padding to mock convolutional layer and to flop calculation

vedant-a-joshi:
This makes sure that the proper output height and width are used in order to determine the proper number of flops in a Convolutional Layer, and adds a test to make sure the implementation accounts for padding. The test explicitly uses the formula.

Tested using `pytest 14_profiling/`, and all tests pass. Also, `tito module test 14` shows that Inline Unit Tests and Module Tests pass.

### DIFF
--- a/tinytorch/src/14_profiling/14_profiling.py
+++ b/tinytorch/src/14_profiling/14_profiling.py
@@ -1322,7 +1322,16 @@ def test_unit_count_conv_flops():
     assert flops_s2 == expected_s2, f"Expected {expected_s2}, got {flops_s2}"
     print(f"✅ Conv2d(3, 64, 7, stride=2): {flops_s2} FLOPs")
 
-    # Test 3: Missing attributes returns 0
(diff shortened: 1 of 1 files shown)


Shashank-Tripathi-07:
Hi @vedant-a-joshi, the only failing check is codespell, and it is not caused by anything in this PR.

Can you update this branch against `dev` (the "Update branch" button on this PR, or `git merge origin/dev` and push)? 

Once the CI is green we can merge it with confidence.

profvjreddi:
Thanks @vedant-a-joshi, and good catch. Integrated into dev.

Two things you couldn't see from your side. The codespell failure on this PR isn't yours, your branch base is about 380 commits behind dev and the file you touched is clean. And I checked your output dims against a real forward pass through Conv2d(3, 16, 3, padding=1) before landing it.

Applied onto a review branch off current dev rather than merging the PR head, so this will show as closed instead of merged. Authorship and credit are intact.

PR #2136, opened 2026-09-15
Fix invalid escape sequences in book tooling scripts

MBK-fr:
### Problem

`binder/tools/scripts/testing/tikz_style_linter.py` (`\}`) and `binder/tools/scripts/content/fix_capitalization.py` (`\i`) use invalid escape sequences in non-raw string literals, so Python 3.12+ raises `SyntaxWarning: invalid escape sequence`.

### Why it matters

The book tooling declares `requires-python >=3.9` and runs the 3.10-3.13 CI matrix; the warnings surface on the 3.12/3.13 jobs.

### Fix

Escape the backslashes; each string value is byte-for-byte identical (ruff W605-style).

### Verification

In-memory `compile()` under `-W error` is clean on both files after (raised before); ASTs unchanged. Targets the `dev` branch.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014HBQNzAf5C2E3MiJC48HQw


profvjreddi:
Thanks, clean fix. I checked the ASTs are identical and these were the only two in binder. Merging.

